PR #131, opened 2026-02-20
feat: configurable status column names

### LINKED ISSUE #53: Configurable status column names in .orchestrator.yml
Replace the fuzzy matching in \`auto_detect_status\` with explicit configuration in \`.orchestrator.yml\`.

## Current behavior

\`scripts/init.sh\` — \`auto_detect_status()\` (around line 170+) queries the GitHub Project's status field options and tries to match them to orchestrator statuses using fuzzy patterns:

```bash
# Current matching logic (simplified):
for option in Backlog "To Do" Todo New; do
  if match "$option" in project_options; then
    status_map[new] = option_id
  fi
done
```

This breaks when projects use non-standard column names like "Triage", "Planned", "Waiting", etc.

## Key files

- \`scripts/init.sh\` — \`auto_detect_status()\` function, around line 170-230
- \`scripts/gh_push.sh\` — \`sync_project_status()\` uses the status map to update project items
- \`.orchestrator.yml\` — per-project config, already has \`gh.repo\` and \`gh.sync_label\`
- \`~/.orchestrator/config.yml\` — global config, stores \`gh.project_status_field_id\` and option IDs

### 1. Add status_map to .orchestrator.yml schema

```yaml
gh:
  repo: "owner/repo"
  project:
    status_map:
      new: "Backlog"        # or "Triage", "New", "Todo"
      in_progress: "In Progress"
      needs_review: "Review"  # or "In Review", "Needs Review"  
      done: "Done"           # or "Closed", "Complete"
      blocked: "Blocked"     # optional
```

### 2. Update init.sh

During \`orch init\` or \`orch project add\`:
- Fetch project status field options (existing code)
- If \`status_map\` is configured, validate that each value exists in project options
- If not configured, fall back to \`auto_detect_status()\` (existing fuzzy matching)
- Store resolved option IDs in config

### 3. Update sync_project_status in gh_push.sh

Instead of looking up option IDs by hardcoded names, read from the stored status map:

```bash
sync_project_status() {
  local task_id="$1" status="$2"
  local option_id
  option_id=$(yq -r ".gh.project_status_options.$status // \"\"" "$CONFIG_PATH")
  [ -n "$option_id" ] || return 0
  # ... existing updateProjectV2ItemFieldValue mutation
}
```

### 4. Interactive setup in init.sh

When \`status_map\` is not configured and auto-detect fails:
```
Could not auto-detect status columns. Available options:
  [1] Triage
  [2] Working
  [3] Review
  [4] Shipped

Map "new" status to: [1]
Map "in_progress" status to: [2]
Map "needs_review" status to: [3]
Map "done" status to: [4]
```

## Tests to add

- \`init.sh uses explicit status_map when configured\`
- \`init.sh falls back to auto_detect when no status_map\`
- \`init.sh validates status_map values against project options\`
- \`sync_project_status uses configured option IDs\`
- \`interactive init prompts for column mapping when auto-detect fails\`

gabrielkoerich:
## Summary
- Add configurable status column names in `.orchestrator.yml` for GitHub Projects V2

Closes #53

### DIFF
--- a/.orchestrator.example.yml
+++ b/.orchestrator.example.yml
@@ -9,7 +9,22 @@
 #   repo: "myorg/myproject"
 #   project_id: "PVT_..."
 #   sync_label: ""
+#   project:
+#     status_map:
+#       new: "Backlog"
+#       in_progress: "In Progress"
+#       needs_review: "Review"
+#       done: "Done"
+#       blocked: "Blocked"   # optional
 #   project_status_field_id: ""
+#   project_status_options:
+#     new: ""
+#     routed: ""
+#     in_progress: ""
+#     blocked: ""
+#     needs_review: ""
+#     in_review: ""
+#     done: ""
 #   project_status_map:
 #     backlog: ""
 #     in_progress: ""
--- a/README.md
+++ b/README.md
@@ -573,6 +573,8 @@ All runtime configuration lives in `config.yml`.
 | `gh` | `sync_label` | Only sync tasks/issues with this label (empty = all). | `"sync"` |
 | `gh` | `project_id` | GitHub Project v2 ID. | `""` |
 | `gh` | `project_status_field_id` | Status field ID in Project v2. | `""` |
+| `gh.project` | `status_map` | Optional mapping from orchestrator statuses to project option names (`new/in_progress/needs_review/done`, `blocked` optional). | `{}` |
+| `gh` | `project_status_options` | Resolved Project option IDs keyed by task status (`new`, `in_progress`, `needs_review`, etc.). | `{}` |
 | `gh` | `project_status_map` | Mapping for `backlog/in_progress/review/done` option IDs. | `{}` |
 | `gh.backoff` | `mode` | Rate-limit behavior: `wait` or `skip`. | `"wait"` |
 | `gh.backoff` | `base_seconds` | Initial backoff duration in seconds. | `30` |
@@ -702,6 +704,21 @@ Project fields belong in `config.yml`:
 gh:
   project_id: ""
   project_status_field_id: ""
+  project:
+    status_map:
+      new: "Backlog"
+      in_progress: "In Progress"
+      needs_review: "Review"
+      done: "Done"
+      blocked: "Blocked" # optional
+  project_status_options:
+    new: ""
+    routed: ""
+    in_progress: ""
+    blocked: ""
+    needs_review: ""
+    in_review: ""
+    done: ""
   project_status_map:
     backlog: ""
     in_progress: ""
@@ -766,6 +783,8 @@ The backoff is shared across pull/push/comment/project updates, so a single rate
 Provide in `config.yml`:
 - `gh.project_id`
 - `gh.project_status_field_id`
+- `gh.project.status_map` (optional explicit option names in `.orchestrator.yml`)
+- `gh.project_status_options` (resolved option IDs by task status)
 - `gh.project_status_map` (Backlog/In Progress/Review/Done option IDs)
 
 #### Finding IDs
--- a/config.example.yml
+++ b/config.example.yml
@@ -48,6 +48,14 @@ gh:
   sync_label: "sync"
   project_id: ""
   project_status_field_id: ""
+  project_status_options:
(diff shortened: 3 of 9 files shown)


gabrielkoerich:

## Summary

The PR is well-structured and the core idea is sound. Backward compatibility is preserved correctly via the two-tier fallback in `project_status_option_id`. However, there are a few issues that should be fixed.

---

### Bug: Double-quote injection in `status_option_id_by_name` (init.sh)

The sed escaping at the top of `status_option_id_by_name` (added around line 137 in `scripts/init.sh`):

```bash
escaped_name=$(printf '%s' "$option_name" | sed -E 's/[][(){}.^$*+?|\\]/\\&/g')
printf '%s' "$status_json" \
  | yq -r ".data.node.fields.nodes[] | select(.name == \"Status\") | .options[] | select(.name | test(\"(?i)^${escaped_name}$\")) | .id"
```

The sed pattern escapes `[ ] ( ) { } . ^ $ * + ? | \` but does **not** escape `"` (double quote). If a GitHub project column name contains a double quote (e.g. `"In Progress"`), the yq expression becomes malformed:

```
select(.name | test("(?i)^\"In Progress\"$"))  # broken string literal
```

This breaks yq parsing entirely. Fix by using a yq variable instead of string interpolation:

```bash
yq -r --arg name "$option_name" \
  '.data.node.fields.nodes[] | select(.name == "Status") | .options[] | select(.name | ascii_downcase == ($name | ascii_downcase)) | .id' \
  2>/dev/null | head -n1
```

This avoids injection entirely and handles case-insensitive matching more robustly.

---

### Bug: `configured_result` initialised to 2 but value from `return 2` is not propagated correctly

In `auto_detect_status` (`scripts/init.sh`):

```bash
local configured_result=2
if apply_configured_status_map "$config_file" "$status_json" "$status_field_id"; then
  return 0
else
  configured_result=$?
fi
if [ "$configured_result" -eq 1 ]; then
  echo "Configured status map validation failed. Falling back to auto-detection." >&2
fi
```

This is actually correct — return 0 → early exit, return 1 → validation failed message, return 2 → not configured (no message, falls through). However, the intent is non-obvious and has no comment. The silent swallow of exit code 2 is fine but should have a brief comment to explain the three-state contract (`0`=used, `1`=failed, `2`=not configured).

---

### Missing temp directory cleanup in three new tests (`tests/orchestrator.bats`)

The three new tests — `init.sh uses configured gh.project.status_map when provided`, `init.sh validates configured status_map values and falls back to auto-detect`, and `interactive status mapping prompts and stores selected options` — all allocate `INIT_DIR=$(mktemp -d)` but never clean up with `rm -rf "$INIT_DIR"`. The existing `init.sh auto_detect_status` test does have cleanup. This is inconsistent and leaks temp directories on the test runner.

---

### Correctness concern: fragile `sed`-based function extraction in interactive test

In `tests/orchestrator.bats`, the interactive test extracts function bodies using:

```bash
$(sed -n '/^save_resolved_status_ids()/,/^}/p' "${REPO_DIR}/scripts/init.sh")
$(sed -n '/^prompt_status_mapping()/,/^}/p' "${REPO_DIR}/scripts/init.sh")
```

The pattern `/^}/` stops at the **first line starting with `}`** after the function start. The `prompt_status_mapping` function contains a nested `pick_option_index()` function. The closing `}` of `pick_option_index` at indent level 2 won't match `^}`, but this extraction is still fragile — any future refactoring that places a closing brace at column 0 inside these functions will silently truncate extraction and break the test. Consider sourcing the whole `init.sh` and calling the function directly, or at minimum add a comment flagging the sensitivity.

---

### Minor: config nesting inconsistency

The user-facing config key `gh.project.status_map` is nested one level deeper than its siblings `gh.project_status_options` and `gh.project_status_field_id`, which sit flat under `gh`. This is a deliberate separation of human-readable names vs. resolved IDs but is not explained anywhere in docs or comments. Adding a brief inline comment in `config.example.yml` / `README.md` distinguishing "human names (`.orchestrator.yml`)" from "resolved IDs (`config.yml`)" would reduce confusion.

---

PR #133, opened 2026-02-20
fix: implement retrospective improvements

gabrielkoerich:
## Summary
- Implement retrospective improvement recommendations
- Optimize gh_push to only sync dirty tasks
- Add tests for retrospective improvement changes

Closes #46

### DIFF
--- a/scripts/db.sh
+++ b/scripts/db.sh
@@ -612,6 +612,17 @@ db_all_task_ids() {
   db_scalar "SELECT id FROM tasks ORDER BY id;"
 }
 
+# Get task IDs that need GitHub sync (dirty content OR stale project-board status).
+# Replaces db_all_task_ids in gh_push to avoid iterating every task on each tick.
+db_syncable_task_ids() {
+  db_scalar "SELECT id FROM tasks
+    WHERE (gh_synced_at IS NULL OR gh_synced_at != updated_at
+           OR gh_issue_number IS NULL
+           OR gh_synced_status IS NULL OR gh_synced_status != status)
+      AND NOT (status = 'done' AND gh_synced_at = updated_at AND (gh_synced_status = 'done' OR gh_synced_status IS NULL))
+    ORDER BY id;"
+}
+
 # Get task count.
 db_total_task_count() {
   db_scalar "SELECT COUNT(*) FROM tasks;"
--- a/scripts/gh_push.sh
+++ b/scripts/gh_push.sh
@@ -212,14 +212,8 @@ if [ "$TASK_COUNT" -le 0 ]; then
   exit 0
 fi
 
-DIRTY_COUNT=$(db_dirty_task_count)
-# Always run the loop if project board is configured (status may need syncing)
-if [ "$DIRTY_COUNT" -le 0 ] && [ -z "$PROJECT_ID" ]; then
-  exit 0
-fi
-
-ALL_IDS=$(db_all_task_ids)
-[ -z "$ALL_IDS" ] && exit 0
+SYNC_IDS=$(db_syncable_task_ids)
+[ -z "$SYNC_IDS" ] && exit 0
 
 while IFS= read -r ID; do
   [ -n "$ID" ] || continue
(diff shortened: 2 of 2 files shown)


gabrielkoerich:

## Code Review: PR #133 - fix: implement retrospective improvements

### Overall
The PR correctly implements the described retrospective improvements: deduplicating system.md rules, guarding SKILLS_DOCS, expanding git diff context, and replacing `db_all_task_ids` with a smarter `db_syncable_task_ids`. The prompt changes and test additions are clean. However, there is a correctness bug in the new SQL query that can cause tasks to be permanently skipped from GitHub sync.

---

### Bug: `db_syncable_task_ids` exclusion clause is too aggressive (`scripts/db.sh`, lines ~619-621)

The AND NOT condition:

```sql
AND NOT (status = 'done' AND gh_synced_at = updated_at AND (gh_synced_status = 'done' OR gh_synced_status IS NULL))
```

The `OR gh_synced_status IS NULL` arm of the exclusion is wrong. Consider a task with:
- `status = 'done'`
- `gh_synced_at = updated_at` (content timestamps match)
- `gh_issue_number = NULL` (no GitHub issue ever created)
- `gh_synced_status = NULL` (status never pushed)

The WHERE clause correctly selects this task (both `gh_issue_number IS NULL` and `gh_synced_status IS NULL` are true). But the AND NOT clause then *excludes* it, because `status = 'done'`, `gh_synced_at = updated_at`, and `gh_synced_status IS NULL` are all true. The result: a 'done' task with no GitHub issue and no synced status is **permanently skipped** by the sync loop.

**Fix**: drop the `OR gh_synced_status IS NULL` from the exclusion:

```sql
AND NOT (status = 'done' AND gh_synced_at = updated_at AND gh_synced_status = 'done')
```

This only skips tasks that are verifiably done and fully synced.

---

### Minor: `build_git_diff` doubles the subshell directory change (`scripts/lib.sh`, lines ~664-669)

```bash
(cd "$dir" && git diff --stat HEAD 2>/dev/null | head -50)
echo ""
(cd "$dir" && git diff HEAD 2>/dev/null | head -200)
```

Two separate `cd "$dir"` subshells are created. This is functionally correct but slightly wasteful. A single subshell would be cleaner:

```bash
(cd "$dir" && { git diff --stat HEAD 2>/dev/null | head -50; echo ''; git diff HEAD 2>/dev/null | head -200; })
```

Not a blocker, but worth noting.

---

### Minor: `build_git_diff` `head -200` increases token cost

Including up to 200 lines of raw diff in agent prompts is intentional per the PR description, but this can meaningfully increase costs for tasks with large uncommitted changesets. The limit is reasonable; just ensure it is the intended trade-off.

---

### Tests (`tests/orchestrator.bats`, lines ~4800-4828)

The tests correctly validate each change. One brittleness note: the exact-count assertions (`[ "$output" -eq 1 ]`) will fail if the patterns are added elsewhere in the files for legitimate reasons in the future. This is acceptable for enforcing the no-duplication invariant, but worth noting in a comment.

Note: the test at line ~4824 (`build_git_diff includes actual diff content not just stat`) checks that `git diff.*--stat` and `git diff HEAD` both appear in `scripts/lib.sh`, which passes with the current implementation. This is a valid smoke test.

---

### Summary

| File | Issue | Severity |
|---|---|---|
| `scripts/db.sh` `db_syncable_task_ids()` | AND NOT excludes tasks with `gh_issue_number IS NULL` and `gh_synced_status IS NULL` when status is 'done' | **Must fix** |
| `scripts/lib.sh` `build_git_diff` | Double `cd` subshell is slightly wasteful | Nit |
| `tests/orchestrator.bats` | Exact-count grep assertions are brittle to future changes | Nit |

The SQL bug is the only blocker. Everything else is clean.

---

gabrielkoerich:
Closing: superseded by the GitHub-native backend migration. The files this PR modifies (db.sh, gh_push.sh) are being deleted and replaced by backend.sh + backend_github.sh.

PR #136, opened 2026-02-20
feat: prevent retry loops on environment failures

### LINKED ISSUE #125: Prevent retry loops on environment/tooling failures
When a task fails with 'command not found' or 'No such file or directory' for a build tool, the orchestrator should not retry automatically. These errors won't resolve without manual intervention. Implementation: 1) In run_task.sh post-processing, detect env failure patterns in agent output (command not found, No such file or directory for known tools) 2) Mark such tasks as needs_review instead of allowing auto-retry 3) Add the detected missing tool to the error message for clarity. On 2026-02-19, 14 tasks were bulk-retried at 02:21 UTC and all immediately failed with the same env issue, wasting tokens.

gabrielkoerich:
## Summary
- Detect environment/tooling failures and prevent retry loops
- Add daily retrospective for 2026-02-19

Closes #125

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -812,6 +812,64 @@ mark_needs_review() {
   append_history "$task_id" "needs_review" "$note"
 }
 
+# Detect environment/tooling failures that won't resolve without manual intervention.
+# Scans text for patterns like "command not found", "No such file or directory" for tools.
+# Returns 0 (success) if an env failure is detected, 1 otherwise.
+# Sets ENV_FAILURE_TOOL to the detected missing tool name (or "unknown").
+# Usage: detect_env_failure "combined stderr+stdout text"
+ENV_FAILURE_TOOL=""
+detect_env_failure() {
+  local text="$1"
+  ENV_FAILURE_TOOL=""
+
+  # Pattern: "<tool>: command not found"
+  local tool
+  tool=$(printf '%s' "$text" | rg -oP '(\S+):\s+command not found' --replace '$1' 2>/dev/null | head -1)
+  if [ -n "$tool" ]; then
+    ENV_FAILURE_TOOL="$tool"
+    return 0
+  fi
+
+  # Pattern: "command not found: <tool>"
+  tool=$(printf '%s' "$text" | rg -oP 'command not found:\s*(\S+)' --replace '$1' 2>/dev/null | head -1)
+  if [ -n "$tool" ]; then
+    ENV_FAILURE_TOOL="$tool"
+    return 0
+  fi
+
+  # Pattern: "No such file or directory" with a recognizable tool path
+  # e.g., "/usr/local/bin/bun: No such file or directory" or "env: node: No such file or directory"
+  tool=$(printf '%s' "$text" | rg -oP '(?:^|[\s/])(\w+):\s+No such file or directory' --replace '$1' 2>/dev/null | head -1)
+  if [ -n "$tool" ]; then
+    ENV_FAILURE_TOOL="$tool"
+    return 0
+  fi
+
+  # Pattern: "env: <tool>: No such file or directory"
+  tool=$(printf '%s' "$text" | rg -oP 'env:\s+(\S+):\s+No such file or directory' --replace '$1' 2>/dev/null | head -1)
+  if [ -n "$tool" ]; then
+    ENV_FAILURE_TOOL="$tool"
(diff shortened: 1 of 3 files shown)


gabrielkoerich:

## Review of PR #136 — feat: prevent retry loops on environment failures

### Critical Bug — Permanent Block After Manual Retry

**`scripts/retry_task.sh` + `scripts/run_task.sh:280`**

`retry_task.sh` resets `status=new` and `agent=NULL` but **does not clear `last_error`**. The `[env]`-tagged string persists in the database. When `run_task.sh` subsequently picks up the task, the guard at ~line 280:

```bash
if [ "$ATTEMPTS" -ge 2 ] && is_env_failure_error "$TASK_LAST_ERROR"; then
  mark_needs_review ...
  exit 0
fi
```

...immediately re-blocks the task to `needs_review` because `last_error` still contains `[env]`. The result is an infinite no-op loop: `retry_task.sh` resets → `run_task.sh` re-blocks → repeat. There is no way for a human to retry after actually fixing the environment without manually editing the database.

**Fix**: In `retry_task.sh`, clear `last_error` when resetting, e.g.:
```bash
db_task_update "$TASK_ID" "status=new" "agent=NULL" "last_error=NULL"
```
This forces `run_task.sh` to re-detect the failure on the next run rather than blocking unconditionally on stale state.

---

### Medium Bug — Pattern 4 is Dead Code + Wrong Tool Name for `env:` Errors

**`scripts/lib.sh` — `detect_env_failure()`, Pattern 3 vs Pattern 4**

Pattern 3 regex: `(?:^|[\s/])(\w+):\s+No such file or directory`

For input `env: solana-test-validator: No such file or directory`, Pattern 3 fires first and captures `env` (the word immediately before the first `: No such file or directory`), setting `ENV_FAILURE_TOOL=env`. Pattern 4 (`env:\s+(\S+):\s+No such file or directory`) — which would correctly capture `solana-test-validator` — is never reached.

The operator message becomes `install env before retrying`, which is misleading.

**Fix**: Reorder so Pattern 4 runs before Pattern 3, or replace Pattern 3 with a more precise regex that excludes the `env:` prefix from the capture group.

---

### Medium Bug — Pattern 3 Generates False Positives

**`scripts/lib.sh` — Pattern 3 in `detect_env_failure()`**

`(\w+):\s+No such file or directory` is too broad. It will match filesystem errors that have nothing to do with missing tools. Examples:

- `fatal: pathspec 'src/main.rs': No such file or directory` → captures `pathspec`
- `open config.json: No such file or directory` → captures `config`
- Any agent that reads a missing file will trigger env-failure classification, causing the task to be wrongly pinned as unretryable.

**Fix**: Tighten Pattern 3 to require a path-like prefix (`/usr/...` or `env:`) before the tool name, so bare filenames are excluded.

---

### Medium Bug — Hyphenated Tool Names Truncated

**`scripts/lib.sh` — Pattern 3 in `detect_env_failure()`**

`\w+` matches only `[a-zA-Z0-9_]`. For `solana-test-validator: No such file or directory`, Pattern 3 captures only `validator`. The error message becomes `install validator before retrying` instead of `install solana-test-validator`. Pattern 1 (`\S+`) handles hyphenated names correctly for `command not found` messages, but Pattern 3 does not.

---

### Low — Pattern 1 Captures Full Absolute Paths

**`scripts/lib.sh` — Pattern 1 in `detect_env_failure()`**

For `/usr/local/bin/bun: command not found`, `\S+` captures the entire absolute path `/usr/local/bin/bun`. The error message becomes `install /usr/local/bin/bun before retrying`. The `[env]` sentinel still functions correctly so this is purely cosmetic, but a `basename` extraction would improve clarity.

---

### Minor — Retrospective References Future/Non-Public Model IDs

**`docs/content/posts/retrospective-2026-02-19.md`**

The table at line 45 references `gpt-5.2` and `gpt-5-codex` as model IDs used for completed tasks. These do not correspond to any publicly available models as of the PR date and may need to be corrected or redacted before publishing.

---

### Summary

The core detection logic (`detect_env_failure`, `is_env_failure_error`) and the automatic-retry prevention are well-structured. The critical issue is that `retry_task.sh` must clear `last_error` when resetting a task — without this, legitimate retries after a human fixes the environment are permanently blocked. The Pattern 3/4 ordering also produces misleading tool names for `env:`-style errors, which affects the quality of operator feedback.

---

PR #170, opened 2026-02-21
Fix job catch-up when last_run is null

### LINKED ISSUE #163: Fix morning-review job catch-up when last_run is null
## Problem

The `morning-review` job has `last_run: null` in `jobs.yml`, which means it **never fires** unless the orchestrator happens to be running at exactly 8:00 AM UTC.

## Root Cause

In `scripts/jobs_tick.sh` lines 53–71:

```bash
if [ -n "$LAST_RUN" ] && [ "$LAST_RUN" \!= "null" ]; then
    # catch-up since last_run — WORKS
elif \! python3 cron_match.py "$SCHEDULE"; then
    continue  # skip if current minute doesn't match — BROKEN
fi
```

When `last_run` is null/empty, the job only fires if the EXACT current minute matches the cron expression. There is no catch-up. The `evening-retrospective` job works because it has a `last_run` value set; `morning-review` has never fired, so `last_run` is perpetually null.

## Fix

When `last_run` is null, treat it as "24 hours ago" to enable catch-up:

```bash
if [ -z "$LAST_RUN" ] || [ "$LAST_RUN" = "null" ]; then
    # First run: use 24h ago as synthetic last_run for catch-up
    LAST_RUN=$(date -u -v-24H +"%Y-%m-%dT%H:%M:%SZ" 2>/dev/null || date -u -d '24 hours ago' +"%Y-%m-%dT%H:%M:%SZ")
fi
if [ -n "$LAST_RUN" ] && [ "$LAST_RUN" \!= "null" ]; then
    # existing catch-up logic...
fi
```

## Expected Result

- `morning-review` job fires after any orchestrator restart if 8am has passed today
- `last_run` gets set on first fire, enabling normal catch-up behavior going forward
- All jobs with null `last_run` behave consistently

## Evidence

```
$ cat ~/.orchestrator/jobs.yml | grep -A5 morning-review
- id: morning-review
  schedule: 0 8 * * *
  enabled: True
  last_run: None   # never fired
  active_task_id: None
```

`jobs.log` has zero entries for morning-review despite days of operation.

gabrielkoerich:
Closes #163

## What changed
- When a job's `last_run` is empty/null, `scripts/jobs_tick.sh` synthesizes a UTC timestamp 24h ago so the existing catch-up logic (`cron_match.py --since`) can detect missed runs.
- Added a bats regression test to ensure a job with `last_run: null` will still catch up for a missed schedule.

## Verification
- `bats tests/orchestrator.bats`

### DIFF
--- a/tests/orchestrator.bats
+++ b/tests/orchestrator.bats
@@ -2454,6 +2454,22 @@ JSON
   [ "$output" = "new" ]
 }
 
+@test "jobs_tick.sh catches up missed job when last_run is null" {
+  # Schedule for 3 hours ago; if last_run is null this should still fire via catch-up.
+  MISSED_HOUR=$(date -u -v-3H +"%H" 2>/dev/null || date -u -d '3 hours ago' +"%H")
+
+  _create_job "test-null-catchup" "Null Catch Up Job" "0 ${MISSED_HOUR} * * *" "task" "Missed job body" "test" "true" "null" "null"
+
+  run env JOBS_FILE="$JOBS_FILE" CONFIG_PATH="$CONFIG_PATH" PROJECT_DIR="$PROJECT_DIR" "${REPO_DIR}/scripts/jobs_tick.sh"
+  [ "$status" -eq 0 ]
+
+  NULL_CATCHUP_ID=$(_task_id_by_title "Null Catch Up Job")
+  [ -n "$NULL_CATCHUP_ID" ]
+  run tdb_field "$NULL_CATCHUP_ID" status
+  [ "$status" -eq 0 ]
+  [ "$output" = "new" ]
+}
+
 @test "jobs_tick.sh does not catch up if last_run is after scheduled time" {
   # last_run is 1 hour ago, schedule is for 3 hours ago — already ran
   LAST_RUN=$(date -u -v-1H +"%Y-%m-%dT%H:%M:%SZ" 2>/dev/null || date -u -d '1 hour ago' +"%Y-%m-%dT%H:%M:%SZ")


gabrielkoerich:

## Review: Fix job catch-up when last_run is null

### Correctness — the fix works

The core logic is sound. In `scripts/jobs_tick.sh` (after the PR's insertion before line 53):

1. **Synthesis block**: when `LAST_RUN` is empty or `"null"`, it is overwritten with a UTC timestamp 24 hours ago.
2. **Dedup guard** (line 53): `[ -n "$LAST_RUN" ] && [ "$LAST_RUN" != "null" ]` is now always `true`, so the dedup and catch-up checks inside always execute.
3. **Dedup check** (lines 55-59): `LAST_RUN_MINUTE` (24h ago) ≠ `NOW_MINUTE_CMP` (current minute), so the job is not skipped — correct.
4. **Catch-up check** (line 63): `cron_match.py "$SCHEDULE" --since "<24h ago>"` fires and detects the missed run — which is exactly the desired behavior.
5. **After task creation**, `last_run` is updated to now, so the next tick uses the real timestamp and the synthesized path is not taken again.

The macOS/Linux `date` fallback pattern (`-v-24H` / `-d '24 hours ago'`) is consistent with the rest of the test file and shell scripts.

---

### One cleanup issue worth noting (not blocking)

**`scripts/jobs_tick.sh` line 68 — dead code**

```bash
elif ! python3 "${SCRIPT_DIR}/cron_match.py" "$SCHEDULE"; then
    # No last_run and current minute doesn't match — skip
    continue
fi
```

This `elif` branch was the old null-last_run guard: it could only be reached when `LAST_RUN` was empty/null (i.e., the `if` at line 53 was false). After the PR's synthesis block, `LAST_RUN` is never empty/null when we reach line 53, so this `elif` is permanently unreachable.

The observable behavior is equivalent — when `cron_match.py --since <24h ago>` returns false AND the current minute doesn't match, the `else continue` at line 65-66 handles the skip. But the `elif` is misleading dead code and should be removed to keep the logic readable.

Suggested cleanup:
```bash
  # Remove lines 68-71:
  elif ! python3 "${SCRIPT_DIR}/cron_match.py" "$SCHEDULE"; then
    # No last_run and current minute doesn't match — skip
    continue
  fi
  # The closing fi at line 71 remains to close the if at line 53
```

This can be done in a follow-up commit and doesn't need to block the PR.

---

### Test coverage — adequate

The new `@test "jobs_tick.sh catches up missed job when last_run is null"` test in `tests/orchestrator.bats` (lines 2457-2472) correctly validates the primary regression: a job with `last_run: null` whose cron was missed 3 hours ago fires via catch-up. The macOS/Linux date compatibility is handled.

One secondary scenario not tested (acceptable for a follow-up): running `jobs_tick.sh` a second time for the same null-last_run job before it completes should not create a duplicate. The existing same-minute dedup logic should handle this, but a test would confirm it.

---

### Summary

The fix is correct, the test covers the regression, and the behavioral change is exactly what the PR description claims. The only actionable item is removing the now-dead `elif` at `scripts/jobs_tick.sh:68-70`, which can be a follow-up.

---

PR #174, opened 2026-02-21

### LINKED ISSUE #153: Auto-reroute tasks to fallback agent on usage limit errors

gabrielkoerich:
## Summary

Auto-reroute tasks when an agent hits a usage/rate limit error, preventing manual intervention and task stalls.

- Detect usage/rate limit errors in agent output (exit codes, stderr, JSON reason field)
- Rotate to the next available agent (codex → claude → opencode), skipping already-tried agents
- Track `limit_reroute_chain` per-task to prevent ping-pong between exhausted agents
- Fall back to `needs_review` only when all agents are exhausted
- Post GitHub issue comments on each reroute for visibility

## Review fixes (addresses feedback from #174 review)

- **Critical**: Only clear `limit_reroute_chain` when `status=done` — not on `in_progress`, which would allow ping-pong to resume on the next iteration
- **High**: Add `local` declarations for `FREE_MODELS`, `_fm`, `_chain` in `reroute_on_usage_limit()` to avoid polluting outer scope
- **Medium**: Use `grep -qF` (fixed-string) in chain exclusion checks in `pick_fallback_agent()` and `reroute_on_usage_limit()`
- **Minor**: Fix `pick_fallback_agent` start index when current agent is not in available list — now correctly starts from `list[0]`
- **Minor**: Remove overly broad `try again later` from `is_usage_limit_error` regex
- **Test**: Add regression test verifying `limit_reroute_chain` survives `in_progress` completions
- **Test**: Update ping-pong stub to use `temporarily unavailable` instead of removed phrase

## Files Changed

- `scripts/lib.sh` — `is_usage_limit_error()`, `pick_fallback_agent()`
- `scripts/run_task.sh` — `reroute_on_usage_limit()` + integration points
- `tests/orchestrator.bats` — 4 tests covering reroute scenarios

Closes #153

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -348,7 +348,7 @@ available_agents() {
 is_usage_limit_error() {
   local text="${1:-}"
   [ -n "$text" ] || return 1
-  printf '%s' "$text" | rg -qi '(?:\b429\b|too many requests|rate[ _-]?limit|usage[ _-]?limit|quota exceeded|insufficient[_ -]?quota|exceeded[_ -]?quota|limit (?:reached|exceeded)|overloaded[_ -]?error|service (?:overloaded|unavailable)|temporarily unavailable|try again later)'
+  printf '%s' "$text" | rg -qi '(?:\b429\b|too many requests|rate[ _-]?limit|usage[ _-]?limit|quota exceeded|insufficient[_ -]?quota|exceeded[_ -]?quota|limit (?:reached|exceeded)|overloaded[_ -]?error|service (?:overloaded|unavailable)|temporarily unavailable)'
 }
 
 # Pick a fallback agent from the locally available agents, rotating after the
@@ -364,7 +364,9 @@ pick_fallback_agent() {
   local list=()
   IFS=',' read -ra list <<< "$agents"
 
-  local start=0
+  local n="${#list[@]}"
+  # Default start so that when current is not found we begin from list[0].
+  local start=$(( n - 1 ))
   for i in "${!list[@]}"; do
     if [ "${list[$i]}" = "$current" ]; then
       start="$i"
@@ -372,7 +374,6 @@ pick_fallback_agent() {
     fi
   done
 
-  local n="${#list[@]}"
   local offset idx candidate
   for offset in $(seq 1 "$n"); do
     idx=$(((start + offset) % n))
@@ -381,7 +382,7 @@ pick_fallback_agent() {
     if [ -n "$current" ] && [ "$candidate" = "$current" ]; then
       continue
     fi
-    if [ -n "$exclude_csv" ] && printf ',%s,' "$exclude_csv" | grep -q ",${candidate},"; then
+    if [ -n "$exclude_csv" ] && printf ',%s,' "$exclude_csv" | grep -qF ",${candidate},"; then
       continue
     fi
     printf '%s' "$candidate"
--- a/scripts/run_task.sh
+++ b/scripts/run_task.sh
@@ -745,7 +745,7 @@ reroute_on_usage_limit() {
   if [ -n "$TASK_AGENT" ]; then
     if [ -z "$chain" ]; then
       chain="$TASK_AGENT"
-    elif ! printf ',%s,' "$chain" | grep -q ",${TASK_AGENT},"; then
+    elif ! printf ',%s,' "$chain" | grep -qF ",${TASK_AGENT},"; then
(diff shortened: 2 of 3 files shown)


gabrielkoerich:

## Code Review: PR #174 — auto-reroute on usage limit errors

### Critical: `limit_reroute_chain` cleared on `in_progress` defeats ping-pong prevention

**File:** `scripts/run_task.sh`, near the end of the added diff (~line 982 post-patch)

```bash
# Clear usage-limit reroute chain on any non-rerouted completion path.
db_task_update "$TASK_ID" "limit_reroute_chain=NULL" 2>/dev/null || true
```

This line runs before checking `AGENT_STATUS`, so it fires on `in_progress` completions too. The failure scenario:

1. Codex hits rate limit → rerouted to claude, chain stored as `codex`
2. Claude processes the task and returns `in_progress` (task not yet done)
3. This line runs → `limit_reroute_chain` is reset to NULL
4. Next iteration: claude also hits a rate limit → chain is empty → `pick_fallback_agent` sees no exclusions → reroutes back to codex
5. Ping-pong resumes, defeating the entire purpose of the chain

**Fix:** Only clear the chain on terminal statuses:
```bash
if [ "$AGENT_STATUS" = "done" ]; then
  db_task_update "$TASK_ID" "limit_reroute_chain=NULL" 2>/dev/null || true
fi
```

---

### High: Missing `local` declarations inside `reroute_on_usage_limit` pollute outer scope

**File:** `scripts/run_task.sh`, `reroute_on_usage_limit()` function

Three variables are assigned inside the function without `local`, making them global:

```bash
FREE_MODELS=$(config_get '.model_map.free // [] | join(",")' 2>/dev/null || true)
# ...
IFS=',' read -ra _fm <<< "$FREE_MODELS"
# ...
IFS=',' read -ra _chain <<< "$chain"
```

`FREE_MODELS` is also used by the auth/billing fallback block later in the same script. `_fm` and `_chain` are array names reused in the outer scope. In practice, the `exit 0` calls prevent actual collision today, but any future refactor that removes early exits would introduce subtle bugs.

**Fix:**
```bash
local FREE_MODELS=""
local -a _fm=()
local -a _chain=()
```

---

### Medium: `grep -q` without `-F` in chain exclusion checks — latent regex injection

**File:** `scripts/lib.sh`, `pick_fallback_agent()`, and `scripts/run_task.sh`, `reroute_on_usage_limit()`

```bash
# lib.sh
if [ -n "$exclude_csv" ] && printf ',%s,' "$exclude_csv" | grep -q ",${candidate},"; then

# run_task.sh
elif ! printf ',%s,' "$chain" | grep -q ",${TASK_AGENT},"; then
```

Both use BRE grep with an agent name interpolated directly into the pattern. Current agent names (`claude`, `codex`, `opencode`) are safe, but this would break if an agent name ever contains `.`, `*`, `[` or other regex metacharacters. Use `grep -qF` for fixed-string matching.

---

### Minor: `pick_fallback_agent` skips `list[0]` when current agent is not in the available list

**File:** `scripts/lib.sh`, `pick_fallback_agent()`

`start` is initialized to `0` and only updated when `current` is found in `list`. The loop iterates `offset in $(seq 1 "$n")`, so the first candidate is always `(start + 1) % n`. When `current` is absent from the list (e.g. decommissioned agent), `start` stays `0` and the first candidate becomes `list[1]`, silently skipping `list[0]`. Low impact but can cause confusing behavior during agent rotation.

---

### Minor: `is_usage_limit_error` pattern `try again later` is overly broad

**File:** `scripts/lib.sh`, `is_usage_limit_error()`

The phrase "try again later" can appear in many non-rate-limit agent error messages (e.g. server error recovery instructions, commit message boilerplate). This could cause spurious reroutes on unrelated errors. Consider anchoring it more tightly, e.g. `(?:please )try again later` or dropping it in favor of the more specific patterns already in the regex.

---

### Tests look good

The three bats tests in `tests/orchestrator.bats` correctly cover:
1. Agent-reported `needs_review` with rate-limit reason → reroute
2. Non-JSON exit-0 output with rate-limit text → reroute  
3. Both agents exhausted → `needs_review` with full chain recorded

The ping-pong prevention test (test 3) would actually expose the `in_progress` chain-clearing bug described above if the stubs returned `in_progress` instead of an error exit — worth adding that scenario to the test suite.

---

gabrielkoerich:
Closing: merge conflicts. Task will be retried.

PR #176, opened 2026-02-21
fix: implement retrospective improvements (P0+P1)

### LINKED ISSUE #148: Review and implement retrospective improvement recommendations

gabrielkoerich:
## Summary

- chore: merge main into gh-task-148-review-and-implement-retrospective-impro
- fix(jobs): use seconds in first-run catch-up timestamp
- fix(run_task): make needs_review guard effective
- fix: implement retrospective improvements (P0+P1)

## Testing

Closes #148

### DIFF
--- a/.orchestrator.example.yml
+++ b/.orchestrator.example.yml
@@ -5,6 +5,14 @@
 # Only include the keys you want to override — everything else
 # falls through to the global config.
 
+# Tools that must be available on PATH before an agent is launched.
+# If any are missing, the task is blocked immediately with a clear reason.
+#
+# required_tools:
+#   - bun
+#   - anchor
+#   - solana-test-validator
+
 # gh:
 #   repo: "myorg/myproject"
 #   project_id: "PVT_..."
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,51 @@
+# Contributing
+
+Thanks for contributing to Orchestrator!
+
+## Development setup
+
+Prereqs (macOS/Homebrew):
+
+```bash
+brew install yq jq just python@3 ripgrep fd bats-core
+# Optional (used in Beads/`bd` experiments): brew install bd
+```
+
+Clone + sanity check:
+
+```bash
+git clone https://github.com/gabrielkoerich/orchestrator.git
+cd orchestrator
+just            # list available commands
+```
+
+## Running tests
+
+Tests use the `bats` framework:
+
+```bash
+bats tests/orchestrator.bats   # main test file
+# bats tests                  # run all tests
+```
+
+## Commit message conventions
+
+Use Conventional Commits:
+- `feat:` new feature (minor bump)
+- `fix:` bug fix (patch bump)
+- `docs:` documentation only
+- `chore:` maintenance/refactor
+
+## PR workflow
+
+- Branch from `main` and keep changes focused.
+- Open a PR early; keep it small and easy to review.
+- Prefer **squash merge** on GitHub.
+
+```bash
+git fetch origin
+git checkout -b my-branch origin/main
+git commit -m "docs: add contributing guide"
+git push -u origin my-branch
+```
+
--- a/README.md
+++ b/README.md
@@ -526,6 +526,7 @@ Place a `.orchestrator.yml` in your project root to override the global config f
 
 ```yaml
 # myproject/.orchestrator.yml
+required_tools: ["bun"]
 gh:
   repo: "myorg/myproject"
   project_id: "PVT_..."
@@ -551,11 +552,14 @@ All runtime configuration lives in `config.yml`.
 | Section | Key | Description | Default |
 | --- | --- | --- | --- |
 | top-level | `project_dir` | Override project directory (auto-detected from CWD). | `""` |
+| top-level | `required_tools` | Tools that must exist on PATH before launching an agent. | `[]` |
 | `workflow` | `auto_close` | Auto-close GitHub issues when tasks are `done`. | `true` |
 | `workflow` | `review_owner` | GitHub handle to tag when review is needed. | `@owner` |
 | `workflow` | `enable_review_agent` | Run a review agent after completion. | `false` |
 | `workflow` | `review_agent` | Fallback reviewer when opposite agent unavailable. | `claude` |
 | `workflow` | `max_attempts` | Max attempts before marking task as blocked. | `10` |
+| `workflow` | `timeout_seconds` | Task execution timeout (0 disables timeout). | `1800` |
+| `workflow` | `timeout_by_complexity` | Per-complexity task timeouts (takes precedence). | `{}` |
 | `workflow` | `required_skills` | Skills always injected into agent prompts (marked `[REQUIRED]`). | `[]` |
 | `workflow` | `disallowed_tools` | Tool patterns blocked via `--disallowedTools`. | `["Bash(rm *)","Bash(rm -*)"]` |
 | `router` | `agent` | Default router executor. | `claude` |
--- a/config.example.yml
+++ b/config.example.yml
@@ -5,8 +5,15 @@ workflow:
   enable_review_agent: false
   review_agent: "claude"
   max_attempts: 10
-  agent_timeout_seconds: 1800  # Max time per agent run (seconds). Increase for complex tasks.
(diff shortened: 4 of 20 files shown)


gabrielkoerich:

## Review of PR #176 — fix: implement retrospective improvements (P0+P1)

### ✅ What looks correct

**`scripts/backend_github.sh:571`** — The `jq` filter correctly excludes `needs_review`-labeled issues from normalization. The `all()` predicate behaves as intended: an issue is excluded if *any* of its labels matches the forbidden set. This is the primary, reliable fix for the needs_review re-run bug.

**`scripts/jobs_tick.sh:68`** — The `--since` flag is properly implemented in `scripts/cron_match.py` (accepts an ISO 8601 timestamp, returns exit 0 if any minute in the window matches the expression). The cross-platform date fallback (`-v-24H` for macOS, `-d '24 hours ago'` for GNU) is correct.

**`scripts/run_task.sh:444` / `scripts/lib.sh:950`** — The timeout refactor is clean. Reading `AGENT_TIMEOUT_SECONDS` from config with env-var override (`${AGENT_TIMEOUT_SECONDS:-$(config_get '...')}`) follows the existing pattern correctly, and the four `tmux_wait` call sites are all updated consistently.

**`prompts/system.md`** — Removing the duplicate rule from the Workflow section is a valid deduplication.

---

### ❌ Bug: `TASK_STATUS` guard in `run_task.sh` is dead code

**`scripts/run_task.sh:122-128` (newly added block)**

```bash
# Guard: never re-run tasks that need human review
if [ "${TASK_STATUS:-}" = "needs_review" ]; then
  log_err "[run] task=$TASK_ID status=needs_review, skipping ..."
  TASK_LOCK_OWNED=false
  exit 0
fi
```

`TASK_STATUS` is not set anywhere in `run_task.sh` before this point. The script queries task fields explicitly via `db_task_field "$TASK_ID" "status"` (around line 88–94 for other checks), but never assigns the result to `TASK_STATUS`. The guard therefore always evaluates against an empty string and **never fires**.

**Fix** — query the status explicitly before the guard:

```bash
TASK_STATUS=$(db_task_field "$TASK_ID" "status")
# Guard: never re-run tasks that need human review
if [ "${TASK_STATUS}" = "needs_review" ]; then
  log_err "[run] task=$TASK_ID status=needs_review, skipping (requires human review before retry)"
  TASK_LOCK_OWNED=false
  exit 0
fi
```

The backend_github.sh normalization exclusion is the real first-line protection, so the system won't actively re-queue `needs_review` tasks. But run_task.sh is also callable directly (e.g., manually, via hooks, or from code paths that bypass normalization), making this guard important. As written it provides only an illusion of safety.

---

### ⚠️ Minor: 24 h catch-up window is silently insufficient for sparse schedules

**`scripts/jobs_tick.sh:68`** — A weekly or monthly job that was never run (fresh install) will only catch up if its most recent firing opportunity falls within the last 24 hours. If the weekly job's cron slot fell 25 hours ago it will silently skip until next week. This is likely acceptable behavior, but the PR description implies the fix is unconditional for null `last_run`. A comment clarifying the intentional 24 h boundary would reduce future confusion.

---

### Summary

The primary concern is the inoperative guard in `run_task.sh` — it must actually read `TASK_STATUS` before the comparison. Everything else is functionally correct.

---

gabrielkoerich:

The changes are logically sound and implement the described fixes correctly. Here are observations, roughly ordered by importance:

**1. Potential lock leak in `run_task.sh:119-124` (minor concern)**
The `needs_review` guard sets `TASK_LOCK_OWNED=false` before `exit 0`. If a task lock is acquired *before* this guard runs, setting `TASK_LOCK_OWNED=false` would tell the EXIT handler not to release it — leaking the lock and potentially blocking future runs of that task. This needs verification against the full `run_task.sh` EXIT handler and lock-acquisition ordering. If the lock is acquired after line 124 (i.e., this guard runs before any lock is held), there is no issue. However, since `backend_github.sh:572` already excludes `needs_review` tasks from normalization, this guard is belt-and-suspenders; the risk is bounded.

**2. `cron_match.py --since` flag not shown in diff (`scripts/jobs_tick.sh:68`)**
The diff truncation means `cron_match.py` changes (if any) are not visible. The new `elif` branch passes `--since "$(date ...)"` to `cron_match.py`, which is the same flag already used in the `if [ -n "$LAST_RUN" ]` branch above it. Since `--since` was pre-existing, this reuse is safe. Worth a note in the PR that `cron_match.py` already supports this.

**3. Date format in `jobs_tick.sh:68`**
The catch-up timestamp uses `+"%Y-%m-%dT%H:%MZ"` (no seconds, literal `Z` suffix). This matches the format apparently expected by `cron_match.py --since` (same format used for `LAST_RUN` elsewhere). Consistent and correct.

**4. macOS/Linux date fallback in `jobs_tick.sh:68`**
The `date -u -v-24H ... 2>/dev/null || date -u -d '24 hours ago' ...` pattern correctly handles both macOS (`-v`) and GNU coreutils (`-d`). No issue.

**5. Redundant `:-1800` defaults in `run_task.sh:457,542,619,660`**
After `AGENT_TIMEOUT_SECONDS` is set from config at line ~444, the subsequent `${AGENT_TIMEOUT_SECONDS:-1800}` fallbacks are redundant. Not a bug — the env var will be set by then — but the code could use `$AGENT_TIMEOUT_SECONDS` directly for clarity. Minor nit.

**6. jq filter correctness in `backend_github.sh:572`**
The updated `all(. != "no-agent" and . != "blocked" and . != "needs_review")` filter is logically correct. Issues with *no labels* still pass (vacuously true), which is existing behavior and intentional.

**7. `system.md` deduplication**
Removing the duplicate "NEVER commit directly to main/master" line from the Workflow section while keeping it in Rules is correct. The retained Rules entry at the top of the file still enforces the constraint.

**Summary**: No blocking issues. The two-layer defense for `needs_review` (exclusion in `db_normalize_new_issues` + guard in `run_task.sh`) is a good approach. The catch-up logic reuses an existing `cron_match.py` capability. The lock concern at point 1 is worth verifying but does not block this PR given the primary exclusion layer in `backend_github.sh`.

---

gabrielkoerich:

The changes are coherent and correctly implement the described fixes. Here are observations per change:

**prompts/system.md**
- The 'NEVER commit directly to main/master' rule was removed. The surrounding context (worktree instructions) implicitly covers this, so the removal is defensible. However, explicit prohibitions are safer in LLM system prompts — an agent operating outside a worktree context would no longer have this guard. Minor risk, likely intentional cleanup.

**scripts/backend_github.sh:571**
- Correctly adds `needs_review` to the normalization exclusion filter, preventing the issue from being picked up and re-assigned during the normalize pass. Comment update is accurate. This pairs well with the run_task.sh guard.

**scripts/jobs_tick.sh:74-84**
- The catch-up logic inversion is correct: instead of 'skip if current minute doesn't match', it now runs if the schedule would have fired in the last 24h. The BSD/GNU date fallback (`-v-24H` vs `-d '24 hours ago'`) is proper cross-platform handling.
- One assumption not verifiable from this diff: `cron_match.py` must already support the `--since <ISO8601>` flag for this to work. If that flag was added in a companion commit not shown here, this is fine; otherwise this would silently do the wrong thing (cron_match.py would likely error or ignore the flag). Worth verifying the flag exists in that script.
- The 24h catch-up window is hardcoded. If jobs are intentionally paused for maintenance longer than 24h, they won't catch up on restart. This is likely acceptable design.

**scripts/lib.sh:973**
- Default `AGENT_TIMEOUT_SECONDS` doubled from 900 to 1800. Consistent with all the `run_task.sh` changes.

**scripts/run_task.sh:119-127**
- The `TASK_STATUS` fallback fetch is correctly guarded with `2>/dev/null || true` to avoid failing if `db_task_field` is unavailable.
- The `needs_review` guard uses `exit 0` (not an error exit), which is the right behavior for a deliberate skip.
- There is a minor TOCTOU window: status is checked at line ~122 but could theoretically change before the task fully runs. For a bash task runner of this type, this is an acceptable best-effort guard.

**scripts/run_task.sh:490, 581, 658, 699**
- All four timeout references consistently changed from 900 to 1800. No missed occurrences visible in this diff.

Overall the logic is sound, the fixes address the stated issues, and there are no security or correctness regressions. The only item worth a quick sanity check before merge is confirming `cron_match.py --since` is supported.

---

PR #178, opened 2026-02-21
feat: trigger tasks from @orchestrator mentions

### LINKED ISSUE #150: @orchestrator mentions in GitHub comments to trigger tasks
Allow tagging @orchestrator in GitHub issue/PR comments to trigger actions. When mentioned, the orchestrator creates a task from the comment context and runs an agent to respond.

gabrielkoerich:
Closes #150

## Summary
- Add `scripts/gh_mentions.sh` to detect actionable `@orchestrator` mentions in issue/PR comments and create a dedicated task issue with captured context.
- Persist comment-id → task-id map in project state for idempotency (no duplicate tasks from the same mention).
- Mirror the agent's structured result comment back onto the originating issue/PR thread (deduped).
- Wire mention polling into `scripts/serve.sh` alongside periodic GitHub work.

## Testing
- `bats tests/orchestrator.bats`

### DIFF
--- a/scripts/gh_mentions.sh
+++ b/scripts/gh_mentions.sh
@@ -3,6 +3,9 @@ set -euo pipefail
 # shellcheck source=scripts/lib.sh
 source "$(dirname "$0")/lib.sh"
 
+# Ensure PROJECT_DIR is set before loading project config (lint + correct state paths)
+PROJECT_DIR="${PROJECT_DIR:-$(pwd)}"
+
 load_project_config
 
 MENTIONS_MARKER="<!-- orch:mention -->"
@@ -62,7 +65,7 @@ acquire_mentions_lock() {
   ensure_state_dir
   local lock_dir="${STATE_DIR}/mentions.lock"
   if mkdir "$lock_dir" 2>/dev/null; then
-    trap 'rmdir "$lock_dir" 2>/dev/null || true' EXIT
+    trap "rmdir \"$lock_dir\" 2>/dev/null || true" EXIT
     return 0
   fi
   return 1
@@ -152,7 +155,7 @@ main() {
     [ -n "$comment_id" ] || continue
     [ -n "$issue_url" ] || continue
 
-    issue_number=$(printf '%s' "$issue_url" | sed -E 's#.*/issues/([0-9]+).*#\\1#')
+    issue_number=$(printf '%s' "$issue_url" | sed -E 's#.*/issues/([0-9]+).*#\1#')
     [ -n "$issue_number" ] || continue
 
     if [ -n "$created_at" ] && [ "$created_at" \> "$max_seen" ]; then
@@ -201,4 +204,3 @@ Acknowledged. Created task #${task_id} from this @orchestrator mention.
 }
 
 main "$@"
-
--- a/tests/orchestrator.bats
+++ b/tests/orchestrator.bats
@@ -1862,7 +1862,7 @@ SH
   [ "$status" -eq 0 ]
 
   # Target issue should have at least the ack + mirrored agent comment
-  run bash -c "jq -r --arg n '${INIT_TASK_ID}' '(.comments[$n] // []) | length' '$GH_MOCK_STATE'"
+  run bash -c "jq -r --arg n '${INIT_TASK_ID}' '(.comments[$n] // []) | length' '$GH_MOCK_STATE' | head -n1"
   [ "$status" -eq 0 ]
   [ "$output" -ge 2 ]
 


gabrielkoerich:

## Review: feat/trigger-tasks-from-orchestrator-mentions

### Bug: Side-effect ordering in `gh_mentions.sh` (lines 178–189)

The acknowledgment comment is posted to GitHub **before** `mentions_db_store` persists the mapping. If the DB write fails (disk full, concurrent writer, permission error), the next run will find no entry for this `comment_id`, create another task, and post another ack — producing duplicate tasks and duplicate ACK comments on the target issue.

Fix: store in DB first, then post the ack. Idempotency state should always be committed before external side effects:

```bash
mentions_db_store "$comment_id" "$task_id" "$issue_number" "${created_at:-}" "${comment_url:-}" || continue
gh_api "repos/${repo}/issues/${issue_number}/comments" -f body="$ack_body" >/dev/null 2>&1 || true
```

### Bug: No pagination — silent mention loss (`gh_mentions.sh` line 142)

```bash
comments_json=$(gh_api -X GET "repos/${repo}/issues/comments" -f since="$since" -f per_page=100 …)
```

Only the first 100 comments since `$since` are fetched. In any busy repo, or after the daemon is offline for a few hours, mentions beyond the first 100 will be silently skipped with no error and no retry. Either add pagination (loop with `?page=N` until the response is empty) or emit a warning when `count` equals 100 so operators know truncation may have occurred.

### Bug: Fragile `issue_number` extraction from URL (`gh_mentions.sh` line 157)

```bash
issue_number=$(printf '%s' "$issue_url" | sed -E 's#.*/issues/([0-9]+).*#\1#')
[ -n "$issue_number" ] || continue
```

If `issue_url` doesn't match the expected pattern, `sed` returns the original string unchanged (e.g. the full URL), and the `[ -n … ]` guard passes with a garbage value. This causes a task to be created with an invalid issue number and an API call to a malformed endpoint. Use a stricter extraction:

```bash
issue_number=$(printf '%s' "$issue_url" | grep -oE '[0-9]+$')
[ -n "$issue_number" ] && [[ "$issue_number" =~ ^[0-9]+$ ]] || continue
```

### Issue: `max_seen` may rewind the `since` cursor on transient errors (`gh_mentions.sh` lines 143–145, 193)

`max_seen` is initialised to the stored `$since` value. If `jq` or `mentions_db_since` fails and returns the epoch default `1970-01-01T00:00:00Z`, then after a successful run `mentions_db_set_since` writes the epoch back, causing the next run to re-scan all comments ever. Add a sanity check: only call `mentions_db_set_since` when `max_seen` is strictly newer than `$since`.

### Issue: Markdown injection via backtick fences in task body (`gh_mentions.sh` lines 102–118)

```bash
\`\`\`markdown
$comment_body
\`\`\`
```

If `comment_body` contains a triple-backtick sequence it will prematurely close the fence, breaking the task body rendering and potentially embedding raw markdown in the task description. Sanitize triple backtick sequences in `comment_body` before embedding (e.g., replace `\`\`\`` with `\`\`\ \``), or use a different quoting mechanism such as indented code (4-space prefix).

### Minor: No deduplication guard on the ACK comment itself

The mirror comment in `backend_github.sh` (lines 1198–1213) correctly deduplicates via a hash stored in `mention_last_mirror_hash`. The initial ACK in `gh_mentions.sh` (lines 179–185) has no such guard. If the `mentions_db_store` is executed but the process dies before writing the lock file (or if there is an exceptional re-entry), a second ACK can appear. Low probability, but inconsistent with the dedup pattern used elsewhere.

### Minor: `seq 0 $((count - 1))` loop is fragile when `count=0`

`[ "$count" -gt 0 ] || return 0` correctly guards this, but if `jq` returns an empty string instead of `0`, the guard fails and `seq 0 -1` produces no output — that is safe, but the defensive check should coerce the value: `count=$(… || echo 0)` (already done) and assert it is numeric before the arithmetic expansion.

### `rg` dependency is fine

`rg` is used throughout the codebase (`run_task.sh`, `lib.sh`, `route_task.sh`, etc.) so the `rg -qi` calls in `gh_mentions.sh` are consistent with project conventions.

### Mock & test changes look correct

The `gh_mock.sh` refactor (adding `owner`/`repo` to comment handlers and the new `_handle_list_repo_issue_comments` endpoint) is correct and the new BATS tests cover the main paths: task creation, idempotency, blockquote filtering, and fenced-code filtering. The mirror assertion `[ "$output" -ge 2 ]` is intentionally loose and sufficient.

### Summary of required changes
1. **`gh_mentions.sh`**: Swap the order of `mentions_db_store` and the ACK `gh_api` call (critical).
2. **`gh_mentions.sh`**: Add pagination or at minimum a truncation warning at 100 results (high).
3. **`gh_mentions.sh`**: Harden `issue_number` extraction to reject non-numeric values (medium).
4. **`gh_mentions.sh`**: Guard `mentions_db_set_since` so it never rewinds to a value older than the current `$since` (medium).
5. **`gh_mentions.sh`**: Sanitize triple-backtick sequences in `comment_body` before embedding (low).

---

PR #181, opened 2026-02-21
fix(github): avoid running needs_review tasks

### LINKED ISSUE #164: Stop orchestrator from re-running needs_review tasks on startup
## Problem

On every orchestrator restart, 9+ issues with `status:needs_review` label are picked up as runnable tasks and crash immediately at `exit=1 at line 1`. This produces noisy log output and wastes resources.

## Affected Issues (as of 2026-02-21)

- #140 (morning review), #142 (CONTRIBUTING.md)
- #126 (dedup system.md), #125 (retry loops), #124 (env validation)
- #53, #51, #50, #49 (older backlog items)

All have `status:needs_review` label and crash within ~2 seconds of being started.

## Log Evidence

```
18:41:20 [run] task=140 starting
18:41:20 [run] task=140 crashed (exit=1) at line 1
18:41:22 [run] task=126 starting
18:41:23 [run] task=126 crashed (exit=1) at line 1
[... 7 more identical crashes ...]
```

## Investigation Needed

1. Why is the GitHub backend including `status:needs_review` issues in the runnable pool?
   - Check `scripts/backend_github.sh` or `scripts/lib.sh` `db_runnable_task_ids()` 
   - `needs_review` should NOT be auto-retried — it requires human attention
2. What specifically causes the crash at line 1?
   - Missing worktree directory? Branch name conflict? Lock file issue?

## Expected Fix

- `needs_review` tasks should never be automatically re-run
- The runnable task query should only return `new` and `routed` status tasks
- If a worktree is missing for a needs_review task, that should not cause a crash

gabrielkoerich:
## Summary

- fix(github): avoid running needs_review tasks

## Changes

```
 CONTRIBUTING.md               |  51 -----------------
 README.md                     |   2 -
 config.example.yml            |   8 ---
 docs/content/cli.md           |   4 +-
 docs/content/configuration.md |   2 -
 docs/content/workflow.md      |   2 -
 scripts/add_task.sh           |  42 +-------------
 scripts/backend_github.sh     |  21 ++++++-
 scripts/create-pr.sh          |  25 ---------
 scripts/gh_project_list.sh    |  51 +++--------------
 scripts/jobs_tick.sh          |   9 ---
 scripts/lib.sh                |  27 ---------
 scripts/make-pr-body.sh       |  36 ------------
 scripts/route_task.sh         |  45 ---------------
 scripts/run_task.sh           |  45 +++++----------
 tests/orchestrator.bats       | 126 ++++++++++--------------------------------
 16 files changed, 73 insertions(+), 423 deletions(-)
```

## Notes

- Base: `main`
- Head: `gh-task-164-stop-orchestrator-from-re-running-needs-`

Closes #164

### DIFF
--- a/scripts/backend_github.sh
+++ b/scripts/backend_github.sh
@@ -538,6 +538,8 @@ db_task_ids_by_status() {
 
   local args=(-f state="$state" -f per_page=100)
   args+=(-f labels="${_GH_STATUS_PREFIX}${_ids_status}")
+  local expected_label="${_GH_STATUS_PREFIX}${_ids_status}"
+  local needs_review_label="${_GH_STATUS_PREFIX}needs_review"
 
   local json
   json=$(gh_api -X GET "repos/$_GH_REPO/issues" "${args[@]}" 2>/dev/null) || return 0
@@ -545,10 +547,23 @@ db_task_ids_by_status() {
   json=$(printf '%s' "$json" | jq -s 'if type == "array" and length > 0 and (.[0] | type) == "array" then [.[][]] else . end' 2>/dev/null || echo "$json")
 
   if [ -n "$exclude_label" ]; then
-    printf '%s' "$json" | jq -r --arg lbl "$exclude_label" \
-      '[.[] | select(.pull_request == null) | select((.labels // []) | map(.name) | all(. != $lbl))] | .[].number' 2>/dev/null || true
+    printf '%s' "$json" | jq -r --arg lbl "$exclude_label" --arg expected "$expected_label" --arg nr "$needs_review_label" '
+      def label_names: (.labels // []) | map(.name);
+      [.[]
+        | select(.pull_request == null)
+        | select(label_names | any(. == $expected))
+        | select(($expected == $nr) or ((label_names | any(. == $nr)) | not))
+        | select(label_names | all(. != $lbl))
+      ]
+      | .[].number' 2>/dev/null || true
   else
-    printf '%s' "$json" | jq -r '.[] | select(.pull_request == null) | .number' 2>/dev/null || true
+    printf '%s' "$json" | jq -r --arg expected "$expected_label" --arg nr "$needs_review_label" '
+      def label_names: (.labels // []) | map(.name);
+      .[]
+      | select(.pull_request == null)
+      | select(label_names | any(. == $expected))
+      | select(($expected == $nr) or ((label_names | any(. == $nr)) | not))
+      | .number' 2>/dev/null || true
   fi
 }
 
--- a/tests/orchestrator.bats
+++ b/tests/orchestrator.bats
@@ -4633,3 +4633,39 @@ SH
   [[ "$output" == *"foo/bar"* ]]
   [[ "$output" == *"$ORCH_HOME/projects/foo/bar.git"* ]]
 }
+
+@test "db_task_ids_by_status excludes issues that also have status:needs_review" {
+  # Create an issue that is "new" but (incorrectly) also has status:needs_review.
+  TASK_OUTPUT=$("${REPO_DIR}/scripts/add_task.sh" "Corrupted status labels" "Should not be runnable" "")
+  TASK2_ID=$(_task_id "$TASK_OUTPUT")
+
+  # Add a second status label without removing the existing one.
+  gh api "repos/${ORCH_GH_REPO}/issues/${TASK2_ID}/labels" -f "labels[]=status:needs_review" >/dev/null
+
+  run bash -c "source '${REPO_DIR}/scripts/lib.sh' && db_task_ids_by_status new"
+  [ "$status" -eq 0 ]
+  [[ \"$output\" != *\"${TASK2_ID}\"* ]]
+}
+
+@test "db_task_ids_by_status still returns needs_review issues when requested" {
+  TASK_OUTPUT=$("${REPO_DIR}/scripts/add_task.sh" "Needs review task" "Should be listed" "")
+  TASK2_ID=$(_task_id "$TASK_OUTPUT")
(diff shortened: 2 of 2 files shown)


gabrielkoerich:

## Core Fix (scripts/backend_github.sh)

The jq logic in `db_task_ids_by_status` is correct. The two new filters work together properly:

- `select(label_names | any(. == $expected))` — client-side confirmation the issue has the queried status label (slightly redundant since the GitHub API already filters by this label, but harmless as a defense against edge cases).
- `select(($expected == $nr) or ((label_names | any(. == $nr)) | not))` — the key fix: when querying for any status other than `needs_review`, issues that *also* carry the `needs_review` label are excluded.

The `else` branch receives the same treatment, so both code paths are consistent.

---

## Critical: Test Assertions Are Broken (tests/orchestrator.bats, ~lines 4503, 4513, 4518, 4526, 4534)

All three new tests use escaped double-quotes in the assertions:

```bash
[[ \"$output\" != *\"${TASK2_ID}\"* ]]
```

This wraps `$output` in *literal* `"` characters and requires the glob pattern to also contain a literal `"` immediately before the task ID. In practice this means:

- The left-hand side evaluates to `"<output_content>"` (e.g., `"163\n164\n165"`).
- The right-hand pattern is `*"164"*`, which requires a literal `"` character directly before `164`.
- For multi-line output the only literal `"` is the opening quote before the *first* ID — so `"164"` never matches any ID that is not the very first line.

**Consequence for the two negative (exclusion) tests**: The test environment always has at least the `INIT_TASK_ID` created in `setup()`. The output of `db_task_ids_by_status` will therefore contain at least two IDs. Because the quoted pattern only matches when the task ID is literally the *first* value in the output, the `!=` assertion passes vacuously — it does **not** verify that the corrupted task is absent. The fix in `backend_github.sh` is not actually validated by these tests.

**Consequence for the positive (inclusion) test** (~line 4518): This would only pass reliably when `TASK2_ID` is the *sole* ID returned for `needs_review`, which is fragile.

The existing convention throughout this file is:
```bash
[[ "$labels" == *"bug"* ]]
[[ "$labels" != *"status:invalid"* ]]
```

Fix all three new assertions to follow the same pattern:
```bash
[[ "$output" != *"${TASK2_ID}"* ]]
[[ "$output" == *"${TASK2_ID}"* ]]
```

---

## Missing Diff Context

The diff is truncated and does not include changes to several files listed in the summary:

- `scripts/lib.sh` (27 deletions) — the tests call `source lib.sh` directly; deletions here could affect test harness behavior.
- `scripts/run_task.sh` (45 changes) — the primary consumer of `db_task_ids_by_status`; verifying the integration of the fix requires seeing this.
- `scripts/jobs_tick.sh` (9 deletions), `scripts/route_task.sh` (45 deletions), `scripts/add_task.sh` (42 deletions).

These cannot be reviewed from the truncated diff. Given the scale of deletions (423 lines removed overall), it is worth confirming nothing related to the fix is inadvertently dropped.

---

## Ancillary Changes (create-pr.sh, make-pr-body.sh)

Both scripts are meaningfully improved:
- `git rev-parse --show-toplevel` is more robust than the old `cd "$(dirname "${BASH_SOURCE[0]}")/.."`.
- Explicit checks for detached HEAD, missing `gh` binary, and zero commits on branch are good defensive additions.
- The regex `gh-task-([0-9]+)($|-)` correctly handles branch names where the issue number is at the end, unlike the old `^gh-task-([0-9]+)-` which would miss such branches.

No issues with these files.

---

gabrielkoerich:

The core logic fix in `scripts/backend_github.sh` is correct, but the tests are broken and will not validate the fix.

## Critical: Test assertions use incorrect quoting (`tests/orchestrator.bats`, lines 4651, 4666, 4681)

All three new tests use backslash-escaped double quotes in the `[[ ]]` assertions:

```bash
[[ \"$output\" != *\"${TASK2_ID}\"* ]]
```

In bash, `\"` inside `[[ ... ]]` is a literal double-quote character. So `\"$output\"` expands to `"<value>"` with literal quote characters surrounding it, and the pattern `*\"${TASK2_ID}\"*` requires literal `"123"` in the output. Since `db_task_ids_by_status` outputs bare numbers (e.g. `123`), the pattern will never match — meaning these `!=` checks **always pass**, even when the bug is present.

Compare with every existing test in the file, e.g. line 4479:
```bash
[[ "$output" == *"foo/bar"* ]]
```
The correct form for the new tests is:
```bash
[[ "$output" != *"${TASK2_ID}"* ]]
```

This affects all three new tests:
- `db_task_ids_by_status excludes issues that also have status:needs_review` (line ~4651)
- `db_task_ids_by_status still returns needs_review issues when requested` (line ~4666)
- `db_task_ids_by_status excludes needs_review issues from routed list when labels are corrupted` (line ~4681)

None of these tests will catch a regression. The fix is trivial — remove the backslashes.

## Logic Review: `scripts/backend_github.sh` lines 541–564

The two new `jq` filters are logically sound:
1. `select(label_names | any(. == $expected))` — defensive check that the issue actually carries the queried label (the GitHub API should already guarantee this, but the guard is harmless).
2. `select(($expected == $nr) or ((label_names | any(. == $nr)) | not))` — correctly passes through issues only when: (a) we're explicitly querying `needs_review`, OR (b) the issue does NOT also carry `status:needs_review`. This is the right guard for corrupted multi-label state.

No issues with the main fix.

## Minor: `needs_review` is hardcoded

The string `needs_review` is hardcoded in `backend_github.sh` line ~542:
```bash
local needs_review_label="${_GH_STATUS_PREFIX}needs_review"
```
This is consistent with how it appears elsewhere in the file, so it's acceptable, but it would be worth a constant/variable if other callers need to reference it. Not a blocker.

---

PR #183, opened 2026-02-21
feat(github): owner slash commands

### LINKED ISSUE #151: Owner feedback commands in GitHub issue comments
Extend the existing owner feedback system to support slash commands in GitHub issue comments.

gabrielkoerich:
## What

Extend the existing owner feedback flow to support repo-owner slash commands in GitHub issue comments (commands are parsed from the **first line**).

Supported commands:
- `/retry`
- `/assign claude|codex|opencode`
- `/unblock`
- `/close`
- `/context <text>` (or put text on following lines)
- `/priority low|medium|high`
- `/help`

## How

- `scripts/poll.sh` scans tasks in `done`, `in_review`, `needs_review`, and `blocked` for new owner comments and applies commands/feedback.
- Non-command owner comments keep existing behavior: append to context + reset to `status:routed`.

## Testing

Closes #151

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -928,7 +928,7 @@ fetch_owner_feedback() {
 
   printf '%s' "$raw" | jq -c \
     --arg owner "$owner_login" --arg since "${since:-}" \
-    '[.[] 
+    '[.[]
       | select(.user.login == $owner)
       # Exclude orchestrator-generated comments (agent response, acks, etc.)
       | select((.body | test("via \\[Orchestrator\\]")) | not)


gabrielkoerich:

Overall the feature is well-structured and the core logic is sound, but there are a few concrete bugs and gaps that should be fixed before merge.

---

## Bugs

### 1. `/help` output renders literal `\n` characters (scripts/lib.sh ~+180)

The help text is passed to `_owner_cmd_ack` as a double-quoted string:
```bash
_owner_cmd_ack "Supported commands:\n\n- `/retry`\n- ...`
```
Inside `_owner_cmd_ack`, the message is interpolated via `printf '%s'` which does **not** interpret `\n` escape sequences. The resulting GitHub comment will contain literal `\n` characters rather than newlines, producing garbled output.

Fix: use `$'...'` ANSI-C quoting so the shell interprets the escapes before they reach printf:
```bash
_owner_cmd_ack $'Supported commands:\n\n- `/retry`\n- `/assign claude|codex|opencode`\n- ...'
```

### 2. `last_error` misused in `/context` handler (scripts/lib.sh ~+155)

```bash
snip=$(printf '%s' "$ctx" | tr '\r\n' '  ' | head -c 200)
...
"last_error=$snip" \
```
`last_error` is semantically an error field; other commands clear it with `last_error=NULL`. Storing a context snippet here will confuse any code or UI that reads `last_error` expecting an error message, and it will survive task completion looking like an error. Either clear it (`last_error=NULL`) or don't touch it at all.

---

## Missing test coverage

### 3. No test for `/close` command (tests/orchestrator.bats)

`/close` is the only state-changing command with no test. It both mutates the task status (`done`) and makes a GitHub API call (`-X PATCH -f state=closed`). Given that the mock infra supports injecting `gh api` responses for other commands, a test verifying `status=done` and that the issue close call is made should be straightforward to add.

### 4. No test for `/priority` command (tests/orchestrator.bats)

The priority→complexity mapping (`high→complex`, `medium→medium`, `low→simple`) has non-obvious behavior; there are also the aliases `med` and the pass-through of `simple|medium|complex` directly. None of this is tested. A test verifying at least one mapping (e.g. `/priority high` sets `complexity=complex` and `status=routed`) would cover the critical path.

### 5. No test for case-insensitivity

The PR description says commands are case-insensitive, and the code does `tr '[:upper:]' '[:lower:]'`, but there is no test exercising e.g. `/RETRY` or `/Assign Claude`.

---

## Minor concerns (non-blocking, but worth a note)

### 6. No lock acquired in `process_owner_feedback_for_task` (scripts/lib.sh ~+186)

The existing `process_owner_feedback` has the comment "Caller must hold the lock." The new wrapper `process_owner_feedback_for_task` is called directly from `poll.sh` without acquiring a task lock. For the states it targets (`done`, `blocked`, `in_review`, `needs_review`) an agent is unlikely to be concurrently writing, but `/retry` and `/unblock` reset to `new`, which agents pick up in the very next poll cycle. If the poll loop itself runs with parallelism this could race. At minimum the comment should document the assumption.

### 7. `advance_ts` set before error-path `db_task_update` calls advance the timestamp even on validation failures

For example, in the `assign` unknown-agent path:
```bash
db_task_update "$task_id" "$advance_ts"
_owner_cmd_ack "❌ Invalid agent..."
```
This is intentional (to avoid re-processing spam), but the behavior is implicit and worth a comment.

---

## Summary

Items 1 and 2 are concrete bugs that will produce incorrect output in production. Items 3–5 are meaningful test gaps for implemented and documented commands. Please fix the `\n` rendering in `/help`, the `last_error` misuse in `/context`, and add tests for `/close` and `/priority` before merging.

---

gabrielkoerich:
@orchestrator fix.

gabrielkoerich:
Closing: shellcheck/semgrep failures. Task will be retried.

gabrielkoerich:
Acknowledged. Created task #202 from this @orchestrator mention.

---
*via [Orchestrator](https://github.com/gabrielkoerich/orchestrator)*

gabrielkoerich:
I dug into the failing CI on this PR. The blocker is shellcheck parse errors (SC1072/SC1073) in scripts/lib.sh where the owner-command ack strings include unescaped markdown backticks (e.g. `/context`, `status:new`). In bash those backticks are command-substitution, which breaks shellcheck and is risky at runtime.

Fix: escape the markdown backticks in the ack/help messages.

I ported the PR changes and applied the fix in a follow-up PR: #207.
- Includes the original owner slash-command feature + docs + tests
- Adds the backtick-escaping fix so shellcheck/security-audit pass
- `HOME=/tmp/orch-test-home bats tests/orchestrator.bats` passes locally

Next step: merge #207 (this PR is closed).

gabrielkoerich:
Acknowledged. Created task #214 from this @orchestrator mention.

---
*via [Orchestrator](https://github.com/gabrielkoerich/orchestrator)*

gabrielkoerich:
Acknowledged. Created task #222 from this @orchestrator mention.

---
*via [Orchestrator](https://github.com/gabrielkoerich/orchestrator)*

PR #185, opened 2026-02-21
fix(jobs): use UTC in cron matching and catch up null last_run jobs

### LINKED ISSUE #144: Investigate morning-review job not firing
The morning-review job has last_run: null and active_task_id: null on 2026-02-19, meaning it never executed. Check: 1) Was the orchestrator server running at 08:00 UTC? 2) Does cron_match.py correctly match '0 8 * * *'? 3) Is the job enabled and properly configured in jobs.yml? 4) Check jobs.log for any matching attempts.

gabrielkoerich:
## Summary

- **Bug 1**: `cron_match.py` called `time.localtime()` instead of `time.gmtime()`, causing `morning-review` to fire at 08:00 local (UTC-3) = 11:00 UTC instead of 08:00 UTC as intended
- **Bug 2**: `jobs_tick.sh` had no catch-up path for jobs with `last_run: null` — when the server was down at the scheduled time and a job had never run before, it was permanently skipped on server restart
- Added 2 new bats tests covering both null last_run scenarios

## Root Cause

The `morning-review` job had `last_run: null` (never run) and the server was down from 2026-02-18T22:00Z to 2026-02-19T20:00Z (~22 hours). Combined bugs:
1. UTC mismatch meant the job would only fire at 11:00 UTC (08:00 local), not 08:00 UTC
2. No catch-up logic for null `last_run` meant the missed schedule was silently dropped on restart

## Test plan
- [x] All 12 `jobs_tick.sh` bats tests pass
- [x] New test: `jobs_tick.sh catches up job with null last_run after server downtime`
- [x] New test: `jobs_tick.sh does not fire null last_run job when no match in last 24h`

Closes #144

### DIFF
--- a/scripts/cron_match.py
+++ b/scripts/cron_match.py
@@ -58,7 +58,7 @@ def matches_cron(expr):
     if len(fields) != 5:
         return False
 
-    now = time.localtime()
+    now = time.gmtime()  # Use UTC for consistency with has_match_since()
     minute, hour, dom, month, dow = fields
 
     # Python tm_wday: 0=Mon..6=Sun → cron dow: 0=Sun, 1=Mon..6=Sat
--- a/scripts/jobs_tick.sh
+++ b/scripts/jobs_tick.sh
@@ -65,9 +65,17 @@ for JOB_ID in $JOB_IDS; do
     else
       continue
     fi
-  elif ! python3 "${SCRIPT_DIR}/cron_match.py" "$SCHEDULE"; then
-    # No last_run and current minute doesn't match — skip
-    continue
+  else
+    # No last_run: check current minute, then 24h catch-up so first-ever
+    # runs aren't silently skipped when the server was down at schedule time.
+    SINCE_24H=$(python3 -c "import datetime; print((datetime.datetime.now(datetime.timezone.utc) - datetime.timedelta(hours=24)).strftime('%Y-%m-%dT%H:%M:%SZ'))")
+    if python3 "${SCRIPT_DIR}/cron_match.py" "$SCHEDULE"; then
+      : # Current minute matches — proceed to execution
+    elif python3 "${SCRIPT_DIR}/cron_match.py" "$SCHEDULE" --since "$SINCE_24H"; then
+      job_log "[jobs] job=$JOB_ID catch-up: no last_run, missed run in last 24h"
+    else
+      continue
+    fi
   fi
 
   if [ "$JOB_TYPE" = "bash" ]; then
--- a/tests/orchestrator.bats
+++ b/tests/orchestrator.bats
@@ -2533,6 +2533,40 @@ JSON
   [ "$output" = "done" ]
 }
 
+@test "jobs_tick.sh catches up job with null last_run after server downtime" {
+  # Simulate a brand-new job (last_run: null) whose scheduled time was missed
+  # because the server was down. The tick should catch up using a 24h lookback.
+  MISSED_HOUR=$(date -u -v-3H +"%H" 2>/dev/null || date -u -d '3 hours ago' +"%H")
+
+  # last_run is null — job has never run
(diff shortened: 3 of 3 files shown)


gabrielkoerich:

Both bugs are correctly identified and fixed. The implementation is logically sound and the tests cover the key scenarios.

## Bug 1: `scripts/cron_match.py` line 61

The fix (`time.localtime()` → `time.gmtime()`) is correct. `has_match_since()` already iterates using `gmtime`, so the two paths were operating in different timezones. This is a clean, correct fix.

## Bug 2: `scripts/jobs_tick.sh` lines 68-76 (new `else` block)

The logic is correct:
- Current-minute match fires immediately (`:` no-op, falls through to execution)
- `--since SINCE_24H` catch-up fires if any match occurred in the last 24 h
- Otherwise `continue` skips as before

The `--since` invocation matches the 4-argument form expected by `cron_match.py` (`argv[1]=schedule, argv[2]=--since, argv[3]=timestamp`), so the interface contract is respected.

**Minor: `SINCE_24H` computed with `python3 -c` inside the loop.** The rest of `jobs_tick.sh` and the tests use the `date -u -v-24H ... || date -u -d '24 hours ago' ...` pattern for cross-platform date arithmetic. Using the same idiom here would be more consistent and avoid an extra python3 subprocess per null-`last_run` job per tick. Example:
```bash
SINCE_24H=$(date -u -v-24H +"%Y-%m-%dT%H:%M:%SZ" 2>/dev/null || date -u -d '24 hours ago' +"%Y-%m-%dT%H:%M:%SZ")
```
This is a style/consistency nit, not a correctness issue — the inline python3 approach is functionally correct and portable.

**Minor: two python3 subprocesses per null-`last_run` job on catch-up ticks.** First call for the current-minute check, second for `--since`. Not a bug, but worth noting for performance on systems with many never-run jobs.

**Minor: only one catch-up fires for multiple missed schedules.** If a job has `schedule: "0 8,10,12 * * *"` and the server was down for 6 hours, only the first missed time causes a catch-up (after which `last_run` is updated past the rest). This is a reasonable design trade-off for the stated use-case, but worth a comment since it may surprise future maintainers.

**The >24h downtime edge case** (`has_match_since` caps lookback at 24 h, and `SINCE_24H` is computed as 24 h ago) means jobs are permanently dropped after very long outages. Already documented inside `has_match_since`; no action needed.

## Tests (`tests/orchestrator.bats` lines 2533–2572)

Both tests are well-constructed.

**Test 1** (`catches up job with null last_run`): Creates a job scheduled at `0 MISSED_HOUR * * *` where `MISSED_HOUR` is 3 hours ago. The cross-platform date syntax (`-v-3H` / `-d '3 hours ago'`) is correct. Since 3 h ago is within the 24 h window and minute 0 of that hour is not the current minute (unless the test happens to run at exactly minute 0 of the same hour, an extremely unlikely race), the `--since` branch will fire and create the task. The assertion on `status=new` is appropriate.

**Test 2** (`does not fire null last_run`): Uses `DOW_2_DAYS = (current_dow + 2) % 7`. For any day of the week, the most recent occurrence of `(dow+2)%7` is exactly 5 days ago, well outside the 24 h window regardless of the hour field being `8`. The `tdb_count` before/after comparison is a solid negative assertion.

Arguments to `_create_job` in both tests map correctly to the helper's positional parameters (`id, title, schedule, type, body, labels, enabled, active_task_id, last_run`), with `last_run="null"` handled by `_create_job`'s `(if $lr == "null" then null else $lr end)` jq expression.

## Summary

All changes are correct. The `localtime`→`gmtime` fix is unambiguously right. The null-`last_run` catch-up logic is sound and the tests provide meaningful coverage. The minor issues above (inline python3 one-liner, two subprocess calls, single-catch-up-per-downtime behavior) are worth a follow-up but do not warrant blocking this PR.

---

PR #187, opened 2026-02-21
feat: owner-only issue filter to prevent task spam

gabrielkoerich:
## Summary
- Only issues from the repo owner (and configured `allowed_authors`) are picked up
- Strangers' issues are silently ignored — no `status:new` label added, not returned by task queries
- Double gate: filtered in `db_normalize_new_issues()` AND `db_task_ids_by_status()`
- Repo owner auto-detected from `gh.repo` config (`owner/repo` → `owner`)

## Config
```yaml
workflow:
  allowed_authors:
    - my-bot
    - trusted-collaborator
```
Repo owner is always allowed, no need to list yourself.

## Test plan
- [x] 5 new tests (owner accepted, stranger rejected, allowed_authors config, query filter, unit tests)
- [x] All 244 tests pass
- [ ] Verify in production: create issue from alt account, confirm it's ignored

### DIFF
--- a/tests/orchestrator.bats
+++ b/tests/orchestrator.bats
@@ -3957,7 +3957,7 @@ SH
   run env PATH="${TMP_DIR}:${PATH}" "${REPO_DIR}/scripts/review_prs.sh"
   [ "$status" -eq 0 ]
   # State file should be empty (no reviews recorded)
-  [ ! -s "${STATE_DIR}/pr_reviews.tsv" ] || [ "$(wc -l < "${STATE_DIR}/pr_reviews.tsv")" -eq 0 ]
+  [ ! -s "${STATE_DIR}/pr_reviews_owner_repo.tsv" ] || [ "$(wc -l < "${STATE_DIR}/pr_reviews_owner_repo.tsv")" -eq 0 ]
 }
 
 @test "review_prs.sh skips already-reviewed PRs at same SHA" {
@@ -3967,7 +3967,7 @@ SH
 
   # Pre-populate review state
   mkdir -p "$STATE_DIR"
-  printf '1\tabc123\tapprove\t2026-01-01T00:00:00Z\tAlready reviewed PR\n' > "${STATE_DIR}/pr_reviews.tsv"
+  printf '1\tabc123\tapprove\t2026-01-01T00:00:00Z\tAlready reviewed PR\n' > "${STATE_DIR}/pr_reviews_owner_repo.tsv"
 
   GH_STUB="${TMP_DIR}/gh"
   cat > "$GH_STUB" <<'SH'
@@ -3995,7 +3995,7 @@ SH
   run env PATH="${TMP_DIR}:${PATH}" "${REPO_DIR}/scripts/review_prs.sh"
   [ "$status" -eq 0 ]
   # Should still have only 1 line in state
-  [ "$(wc -l < "${STATE_DIR}/pr_reviews.tsv")" -eq 1 ]
+  [ "$(wc -l < "${STATE_DIR}/pr_reviews_owner_repo.tsv")" -eq 1 ]
 }
 
 @test "review_prs.sh reviews new PR and records state" {
@@ -4041,7 +4041,7 @@ SH
   [ "$status" -eq 0 ]
 
   # Should have recorded the review
-  run grep "42" "${STATE_DIR}/pr_reviews.tsv"
+  run grep "42" "${STATE_DIR}/pr_reviews_owner_repo.tsv"
   [ "$status" -eq 0 ]
   [[ "$output" == *"def456"* ]]
   [[ "$output" == *"approve"* ]]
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

The PR correctly implements owner-only filtering with a sensible double-gate design. Core security logic is sound and tests cover the critical paths. A few issues worth noting:

**[Bug — performance] Broken cache in `_gh_allowed_authors` (`backend_github.sh:28-41`)**
The `_GH_ALLOWED_AUTHORS` global-variable cache never persists because every caller uses `$()` command substitution, which forks a subshell. Assignments inside a subshell don't propagate to the parent. As a result, in the `db_normalize_new_issues` while loop, each iteration of `_gh_is_allowed_author` (line ~643) spawns a fresh `_gh_allowed_authors` call that re-runs `config_get` from scratch. The feature is *correct* — the function always returns the right list — but the intended O(1) cache hit is never taken; it's O(n × config_get) across unlabeled issues.

Fix: capture `allowed_csv` once before the loop (already done for the jq paths), then pass it as a parameter to `_gh_is_allowed_author` or inline the comparison, rather than calling `_gh_allowed_authors` again inside the loop.

**[Minor — test isolation] Inline cleanup can leave dirty state (`tests/orchestrator.bats:4706-4710`)**
The `db_normalize_new_issues accepts issues from allowed_authors config` test calls `yq -i 'del(.workflow.allowed_authors)'` at the end of the test body. If the test fails before that line (e.g., the label assertion fails), `$CONFIG_PATH` retains `allowed_authors: [trusted-bot]`, which could cause subsequent tests to behave unexpectedly. Move the cleanup to a `teardown` function or use bats's `--no-preserve-state` pattern to guarantee cleanup regardless of failure.

**[Info — open fallback] Both gates fall open on `_gh_ensure_repo` failure (`backend_github.sh:603` and `~648`)**
`allowed_csv=$(_gh_allowed_authors 2>/dev/null || true)` means an empty string is passed when the function fails. Both the jq `if $allowed == "" then true` path and the `if [ -n "$allowed_csv" ]` guard degrade gracefully to pre-PR behavior (all issues allowed). This is consistent and intentional — note it in the PR description so future maintainers understand it is a deliberate safe-open fallback, not an oversight.

**[Info — double-gate] Defense-in-depth is correct**
Filtering in both `db_normalize_new_issues` (preventing the `status:new` label from being applied) and `db_task_ids_by_status` (preventing already-labeled stranger issues from being returned) is the right design. This handles the edge case where issues exist that were labeled before the feature was enabled.

---

gabrielkoerich:

The implementation is logically sound in the happy path and tests cover the main scenarios, but there is one meaningful security issue and a few smaller correctness concerns.

## Critical: Fail-open when repo detection fails

**backend_github.sh, `db_task_ids_by_status` (~line 584) and `db_normalize_new_issues` (~line 624):**

Both call sites capture the allowed-authors list with:
```bash
allowed_csv=$(_gh_allowed_authors 2>/dev/null || true)
```
If `_gh_ensure_repo` fails (network issue, misconfigured `gh.repo`, first-run race), `_gh_allowed_authors` returns 1 and `allowed_csv` is set to the empty string. Both consumers then treat empty as "allow everything":
- In jq: `def is_allowed: if $allowed == "" then true else ...`
- In the normalize loop: `if [ -n "$allowed_csv" ] && ! _gh_is_allowed_author ...`

This means a transient config or network error silently disables the spam filter for that entire invocation. For a feature whose entire purpose is authorization, failing open defeats the point. Suggested fix: propagate the failure or, at minimum, emit a loud warning and skip normalization/task-listing entirely rather than allowing all authors through:
```bash
allowed_csv=$(_gh_allowed_authors) || { log_err "[filter] could not determine allowed authors, skipping"; return 0; }
```

## Medium: Empty `_GH_REPO` yields malformed CSV

**backend_github.sh, `_gh_allowed_authors` (~line 33):**
```bash
local owner="${_GH_REPO%%/*}"
_GH_ALLOWED_AUTHORS="$owner"
```
If `_GH_REPO` is unset or is a bare name without a slash, `owner` could be empty or the full string, and later appends produce a leading comma (`,bot1,bot2`). `_gh_ensure_repo` is called first and presumably validates the format, but if that contract is ever relaxed, this silently produces a malformed list. A one-line guard `[ -n "$owner" ] || return 1` would make this explicit.

## Minor: Redundant double-fetch in `db_normalize_new_issues`

**backend_github.sh, `db_normalize_new_issues` (~line 640):**
`allowed_csv` is computed once for the empty-check guard, then `_gh_is_allowed_author` is called inside the loop — which itself calls `_gh_allowed_authors` again (hitting the cache, so no real cost, but it's confusing). The pre-fetch of `allowed_csv` is only used for `[ -n "$allowed_csv" ]`. This guard is inconsistent with `_gh_is_allowed_author`'s own behavior on failure (it returns 1, not a pass-all). The simplest fix is to remove the pre-fetch and let `_gh_is_allowed_author` own the decision entirely.

## Minor: Cache not invalidated between tests or on config reload

**backend_github.sh, global `_GH_ALLOWED_AUTHORS`:** The process-lifetime cache is appropriate for short-lived CLI invocations. Tests correctly reset it with `_GH_ALLOWED_AUTHORS=""`. But the PR description mentions config-driven `allowed_authors` — if an operator adds a new trusted author to the YAML while the orchestrator is running, the change won't take effect until restart. This is acceptable behavior but worth a comment in `_gh_allowed_authors` so future maintainers understand the tradeoff.

## Tests look solid

The five new tests in `tests/orchestrator.bats` cover the important cases (owner accepted, stranger rejected, `allowed_authors` config, query filter, unit tests). The `_GH_ALLOWED_AUTHORS=""` cache resets are present and correct. The mock extension in `gh_mock.sh` (`GH_MOCK_LOGIN`) is clean and doesn't break existing tests.

## Summary

The fail-open behavior on `_gh_allowed_authors` failure is the only issue that should block merge — it undermines the security goal of the PR. The other items are correctness improvements worth addressing but not blockers on their own.

---

PR #188, opened 2026-02-21

### LINKED ISSUE #153: Auto-reroute tasks to fallback agent on usage limit errors

gabrielkoerich:
## Summary

- Adds `is_usage_limit_error()` helper in `scripts/lib.sh` to detect rate/usage limit errors via regex (429, too many requests, rate-limit, quota exceeded, overloaded, etc.)
- Adds `pick_fallback_agent()` helper that rotates to the next available agent, skipping a configurable exclude list to prevent ping-pong
- Adds `reroute_on_usage_limit()` in `scripts/run_task.sh` that auto-reroutes tasks when usage/rate limits are detected:
  - On non-zero exit code with usage-limit output
  - On invalid/empty JSON response (likely agent crash due to limit)
  - On agent response with missing status field
- Tracks `limit_reroute_chain` per-task to avoid bouncing between agents; chain only clears on `done`
- Posts GitHub issue comment with snippet when rerouting or when no fallback is available
- Extracts auth/billing error path to use `pick_fallback_agent()` for consistency

## Test plan
- [x] 237 bats tests passing including 3 reroute-specific tests
- [x] Auto-reroutes on usage limit (agent-reported needs_review)
- [x] Auto-reroutes on non-JSON response (crash/limit)
- [x] Preserves limit_reroute_chain on in_progress completions (ping-pong prevention)

Closes #153

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -348,7 +348,7 @@ available_agents() {
 is_usage_limit_error() {
   local text="${1:-}"
   [ -n "$text" ] || return 1
-  printf '%s' "$text" | rg -qi '(?:\b429\b|too many requests|rate[ _-]?limit|usage[ _-]?limit|quota exceeded|insufficient[_ -]?quota|exceeded[_ -]?quota|limit (?:reached|exceeded)|overloaded[_ -]?error|service (?:overloaded|unavailable)|temporarily unavailable|try again later)'
+  printf '%s' "$text" | rg -qi '(?:\b429\b|too many requests|rate[ _-]?limit|usage[ _-]?limit|quota exceeded|insufficient[_ -]?quota|exceeded[_ -]?quota|limit (?:reached|exceeded)|overloaded[_ -]?error|service (?:overloaded|unavailable)|temporarily unavailable)'
 }
 
 # Pick a fallback agent from the locally available agents, rotating after the
@@ -364,7 +364,9 @@ pick_fallback_agent() {
   local list=()
   IFS=',' read -ra list <<< "$agents"
 
-  local start=0
+  local n="${#list[@]}"
+  # Default start so that when current is not found we begin from list[0].
+  local start=$(( n - 1 ))
   for i in "${!list[@]}"; do
     if [ "${list[$i]}" = "$current" ]; then
       start="$i"
@@ -372,7 +374,6 @@ pick_fallback_agent() {
     fi
   done
 
-  local n="${#list[@]}"
   local offset idx candidate
   for offset in $(seq 1 "$n"); do
     idx=$(((start + offset) % n))
@@ -381,7 +382,7 @@ pick_fallback_agent() {
     if [ -n "$current" ] && [ "$candidate" = "$current" ]; then
       continue
     fi
-    if [ -n "$exclude_csv" ] && printf ',%s,' "$exclude_csv" | grep -q ",${candidate},"; then
+    if [ -n "$exclude_csv" ] && printf ',%s,' "$exclude_csv" | grep -qF ",${candidate},"; then
       continue
     fi
     printf '%s' "$candidate"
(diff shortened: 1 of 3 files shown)


gabrielkoerich:

Overall the design is sound and the ping-pong prevention via `limit_reroute_chain` is well thought out. Tests cover the main happy paths. However several issues need to be addressed before merging:

## High / Medium Issues

### 1. Overly broad regex patterns — false positives (lib.sh, `is_usage_limit_error`)

Two patterns in the regex will match completely unrelated errors and incorrectly trigger agent rerouting:

- `service (?:overloaded|unavailable)` matches any `503 Service Unavailable` network error (DNS failure, gateway timeout, upstream maintenance). These should be retried or marked `needs_review`, not rerouted.
- `temporarily unavailable` alone is extremely common in local process errors (SSH, filesystem, startup failures) and has no meaningful signal for rate limiting.

Suggested fix: require co-occurrence context or tighten these patterns. For example, `service overloaded` is Anthropic-specific and fine to keep, but `service unavailable` is a generic HTTP phrase that should be dropped or anchored. `temporarily unavailable` should require adjacent rate-limit context.

### 2. Stale `limit_reroute_chain` after task requeue (run_task.sh ~L1031)

The chain is cleared only on `AGENT_STATUS = done`. If a task completes with `done`, has its chain cleared, then is **re-queued** for a new run (e.g., a follow-up delegation), the chain field starts empty — which is correct. But if a task is re-queued while the chain from a *prior* limit event is still in the DB (i.e., the previous run ended with `needs_review` rather than `done`), the stale chain pre-excludes agents that are no longer rate-limited. The chain should also be cleared when `status=new` is written at the start of a fresh task dispatch, not only on `done`.

### 3. Auth/billing model cycling regression (run_task.sh, `ATTEMPTS - 1` change)

The original code counted opencode-specific history entries (`grep -c "switched to opencode"`), so index 0 was always the first free model when opencode was first selected. The replacement `ATTEMPTS - 1` counts *all* task attempts regardless of which agent ran them. If a task reaches the auth/billing handler on attempt 8, it skips to `_fm[7 % len]`, bypassing earlier free models. While the modulo prevents crashes, the model selection is no longer anchored to the opencode-specific switch count. This is a silent behavior change not called out in the PR description. At minimum document the intentional change; ideally use a more semantically correct index.

### 4. Bare `quota` match removed — silent regression (lib.sh regex / run_task.sh ~L869)

The old unified regex matched bare `quota` as a substring. The new `is_usage_limit_error` requires `quota exceeded`, `exceeded[_ -]?quota`, or `insufficient[_ -]?quota`. The trimmed auth/billing regex no longer includes `quota` at all. An API response containing only the word `quota` (e.g., some provider error formats) now falls through to the generic non-zero exit path without either classification. Verify whether any known provider returns bare `quota` in errors; if so, add it back.

### 5. Snippet may expose credentials in GitHub comments (run_task.sh `reroute_on_usage_limit`)

`snippet = tail -c 300` of `combined_output` is posted verbatim to GitHub issue comments. Provider error messages routinely include partial API key identifiers (e.g. `"Invalid key: sk-proj-abc..."`) in the last bytes of stderr. This pre-exists in the auth/billing path, but this PR adds two more callsites. Consider redacting patterns like `sk-[a-zA-Z0-9-]{10,}` or simply omitting the snippet from public comments and keeping it only in the internal `last_error` field.

## Minor Issues

### 6. Missing negative test cases for `is_usage_limit_error`

No BATS tests verify that inputs like a plain `503 Service Unavailable` from a network timeout, or `temporarily unavailable` from a non-rate-limit source, are *not* misclassified. Given the false-positive risk in issue #1, at least one negative-case test should be added.

### 7. PR description says 3 reroute tests, diff adds 4

The test `run_task.sh preserves limit_reroute_chain when agent returns in_progress` is the 4th test. Minor doc inconsistency only.

## What's Good

- `pick_fallback_agent` rotation and deduplication logic is correct; the comma-wrap grep pattern avoids substring match bugs.
- Chain-only-cleared-on-done is the right semantic (not on `in_progress`) and the regression test for it is solid.
- Separating usage-limit detection from auth/billing detection is cleaner than the original monolithic regex.
- The three existing reroute tests cover the core paths well (agent-reported, non-JSON, and exhaustion).

---

PR #189, opened 2026-02-22
fix: jobs_tick null last_run catch-up, needs_review guard, 1800s timeout

### LINKED ISSUE #140: Daily morning review: improvements and optimization
Morning check-in for orchestrator maintenance and optimization.

FIRST: Look at the most recent evening retrospective task (label: job:evening-retrospective, status: done).
Read its summary to see what was identified yesterday — carry forward any unfinished items or priorities.

Then check:
1) Are any tasks stuck or failing repeatedly? Diagnose and fix root causes.
2) Are there test gaps or flaky tests? Add missing tests.
3) Are there error patterns in logs (~/.orchestrator/.orchestrator/orchestrator.log)?
4) Can any scripts be simplified or optimized?
5) Are docs (docs/content/*.md, README.md) accurate and complete?
6) Check GitHub issues for owner feedback that needs attention.

Prioritize fixes that improve reliability and agent success rate.
If the evening retro flagged specific issues, tackle those first.

### LINKED ISSUE #163: Fix morning-review job catch-up when last_run is null

### LINKED ISSUE #164: Stop orchestrator from re-running needs_review tasks on startup

### LINKED ISSUE #166: Increase default task timeout: 900s is too short for medium-complexity tasks
## Problem

The default task timeout of 900 seconds (15 minutes) is too short for medium-complexity codex tasks. Task #158 (add task count to status output) timed out at exactly 900s despite completing valid work — the PR was created, approved, and merged.

## Evidence

```
18:57:04 [run] task=158 tmux session timed out after 900s
18:57:04 [run] task=158 agent finished (exit=124) duration=15m 23s
18:57:05 [run] task=158 TIMEOUT after 15m 23s
```

Yet the task had already written a valid PR that was reviewed and merged (PR #162). The agent was blocked by the timeout even though it had finished its core work.

## Proposed Fix

Increase the default timeout or make it configurable by complexity tier:

```yaml
# config.yml
workflow:
  timeout_seconds: 1800  # increase from 900 to 1800

# Or per-complexity:
workflow:
  timeout_by_complexity:
    simple: 600
    medium: 1800
    complex: 3600
```

A medium-complexity task on codex typically involves:
1. Read existing code (~1-2 min)
2. Edit files (~2-5 min)
3. Run tests (blocked!) or verify manually (~2 min)
4. Commit (~1 min)
5. Push + create PR (~2-3 min)
6. Total: ~10-15 min — very close to the 900s limit

## Impact

Tasks that hit the timeout get `needs_review` status even when they completed successfully. This adds noise to the backlog.

gabrielkoerich:
## Summary

Morning review fixes for reliability and agent success rate:

- **fix(jobs_tick)**: synthesize 24h lookback when `last_run` is null so catch-up fires correctly on the first scheduler run (fixes #163)
- **fix(run_task)**: guard against re-running `needs_review` tasks at startup — exit 0 immediately without touching the agent (fixes #164)
- **fix(lib,run_task)**: increase default agent timeout from 900s → 1800s in `run_with_timeout` and all `tmux_wait` calls; reads from `config.yml` `workflow.agent_timeout_seconds` (fixes #166)
- **fix(backend_github)**: exclude `needs_review` issues from unlabeled normalization to prevent unwanted status resets
- **docs**: document `agent_timeout_seconds` in README, `config.example.yml`, and `docs/content/configuration.md`

## Test plan

- [x] `bats tests/orchestrator.bats --filter "needs_review"` → new test passes
- [x] `bats tests/orchestrator.bats --filter "jobs_tick"` → all 11 jobs_tick tests pass
- [x] Full test suite clean (no regressions)

Closes #140

### DIFF
--- a/README.md
+++ b/README.md
@@ -556,6 +556,7 @@ All runtime configuration lives in `config.yml`.
 | `workflow` | `enable_review_agent` | Run a review agent after completion. | `false` |
 | `workflow` | `review_agent` | Fallback reviewer when opposite agent unavailable. | `claude` |
 | `workflow` | `max_attempts` | Max attempts before marking task as blocked. | `10` |
+| `workflow` | `agent_timeout_seconds` | Max time (seconds) per agent run. Increase for complex tasks. | `1800` |
 | `workflow` | `required_skills` | Skills always injected into agent prompts (marked `[REQUIRED]`). | `[]` |
 | `workflow` | `disallowed_tools` | Tool patterns blocked via `--disallowedTools`. | `["Bash(rm *)","Bash(rm -*)"]` |
 | `router` | `agent` | Default router executor. | `claude` |
--- a/config.example.yml
+++ b/config.example.yml
@@ -5,6 +5,7 @@ workflow:
   enable_review_agent: false
   review_agent: "claude"
   max_attempts: 10
+  agent_timeout_seconds: 1800  # Max time per agent run. Increase for complex tasks (default: 1800s).
 router:
   agent: "claude"
   model: "haiku"
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -16,6 +16,7 @@ All runtime configuration lives in `~/.orchestrator/config.yml`.
 | `workflow` | `enable_review_agent` | Run a [review agent](@/review-agent.md) after task completion | `false` |
 | `workflow` | `review_agent` | Fallback reviewer when opposite agent unavailable | `claude` |
 | `workflow` | `max_attempts` | Max attempts before marking task as blocked | `10` |
+| `workflow` | `agent_timeout_seconds` | Max time (seconds) per agent run. Increase for complex tasks. | `1800` |
 | `workflow` | `stuck_timeout` | Timeout (seconds) for detecting stuck in_progress tasks | `1800` |
 | `workflow` | `required_skills` | Skills always injected into agent prompts (marked `[REQUIRED]`) | `[]` |
 | `workflow` | `disallowed_tools` | Tool patterns blocked via `--disallowedTools` | `["Bash(rm *)","Bash(rm -*)"]` |
--- a/scripts/backend_github.sh
+++ b/scripts/backend_github.sh
@@ -561,12 +561,12 @@ db_normalize_new_issues() {
   json=$(gh_api -X GET "repos/$_GH_REPO/issues" \
     -f state=open -f per_page=50 -f sort=created -f direction=desc 2>/dev/null) || return 0
 
-  # Find issues (not PRs) without any status: label, excluding no-agent and blocked
+  # Find issues (not PRs) without any status: label, excluding no-agent, blocked, and needs_review
   local unlabeled_ids
   unlabeled_ids=$(printf '%s' "$json" | jq -r --arg p "$_GH_STATUS_PREFIX" '
     [.[] | select(.pull_request == null)
          | select((.labels // []) | map(.name) | all(startswith($p) | not))
-         | select((.labels // []) | map(.name) | all(. != "no-agent" and . != "blocked"))]
+         | select((.labels // []) | map(.name) | all(. != "no-agent" and . != "blocked" and . != "needs_review"))]
     | .[].number' 2>/dev/null || true)
 
   [ -n "$unlabeled_ids" ] || return 0
--- a/scripts/jobs_tick.sh
+++ b/scripts/jobs_tick.sh
@@ -50,23 +50,27 @@ for JOB_ID in $JOB_IDS; do
   fi
 
   # Schedule matching with catch-up for missed runs during downtime
-  if [ -n "$LAST_RUN" ] && [ "$LAST_RUN" != "null" ]; then
-    # Prevent duplicate creation if tick runs multiple times in the same minute
-    LAST_RUN_MINUTE=$(printf '%s' "$LAST_RUN" | cut -c1-16)
-    NOW_MINUTE_CMP=$(date -u +"%Y-%m-%dT%H:%M")
-    if [ "$LAST_RUN_MINUTE" = "$NOW_MINUTE_CMP" ]; then
-      continue
-    fi
+  # If this is the first run (no last_run), synthesize a 24h lookback so catch-up can fire.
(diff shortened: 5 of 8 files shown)


gabrielkoerich:

The core fixes are correct and well-reasoned — the `needs_review` guard, 24h lookback, timeout increase, and `backend_github.sh` jq fix all address real bugs. However there is one concrete test correctness issue that should be fixed before merge.

**[Bug] `tests/orchestrator.bats` — dead assertion in the `needs_review` test (~line 1036)**

The test has:
```bash
# Status must remain needs_review (not changed by agent)
run tdb_field "$SKIP_ID" status
[ "$output" = "needs_review" ]

# Agent must not have been called
[[ "$output" != *"AGENT_WAS_CALLED"* ]]
```

After `run tdb_field "$SKIP_ID" status`, bats replaces `$output` with the status value `"needs_review"`. The final assertion is therefore checking whether the string `"needs_review"` contains `"AGENT_WAS_CALLED"` — which is always `true` regardless of whether the agent ran. The agent-not-called invariant is never actually verified.

Additionally, the stub writes to **stderr** (`echo "AGENT_WAS_CALLED" >&2`), and even if `$output` hadn't been clobbered by the second `run`, the check may still pass vacuously depending on bats version and stderr capture behavior.

The fix is to save the run_task.sh output before overwriting it:
```bash
run env PATH="${TMP_DIR}:${PATH}" ... "${REPO_DIR}/scripts/run_task.sh" "$SKIP_ID"
RT_OUTPUT="$output"  # save before next run overwrites it
[ "$status" -eq 0 ]

run tdb_field "$SKIP_ID" status
[ "$output" = "needs_review" ]

# Agent must not have been called
[[ "$RT_OUTPUT" != *"AGENT_WAS_CALLED"* ]]
```

Or alternatively redirect the stub output to a sentinel file and check for its existence.

---

**[Observation] `scripts/backend_github.sh:569` — `needs_review` bare label exclusion**

The exclusion of bare `needs_review` label is correct as a defense-in-depth measure. Note that in normal operation, the outer filter `all(startswith($p) | not)` (where `$p = "status:"`) already excludes issues with `status:needs_review`. The inner bare exclusion guards against manually-applied labels not using the prefix convention. No change needed, just worth documenting.

---

**[Observation] `scripts/jobs_tick.sh` — date fallback edge case**

If both `date -v-24H` (BSD) and `date -d '24 hours ago'` (GNU) fail, `LAST_RUN` becomes an empty string. The code then passes `--since ""` to `cron_match.py`. This is extremely unlikely in practice and degrades to current-minute-only matching, so it is not blocking — but adding `|| true` or a non-empty guard would make the failure mode explicit.

---

**[Confirmed correct]** All three `tmux_wait` call sites in `run_task.sh` (diff lines +548, +615, +656) are correctly updated from `"${AGENT_TIMEOUT_SECONDS:-900}"` to `"${AGENT_TIMEOUT_SECONDS:-1800}"`. The `AGENT_TIMEOUT_SECONDS` config-reading line is placed at the right point in the file (before the first `tmux_wait` usage). The `lib.sh:950` `run_with_timeout` default change is consistent.

---

PR #190, opened 2026-02-22
feat(github): owner slash commands in issue comments

### LINKED ISSUE #151: Owner feedback commands in GitHub issue comments

gabrielkoerich:
## What

Extend the existing owner-feedback flow to support repo-owner **slash commands** in GitHub issue comments (parsed from the **first line**, case-insensitive).

Commands:
- \/retry
- \/assign claude|codex|opencode
- \/unblock
- \/close
- \/context <text> (or put text on following lines)
- \/priority low|medium|high (or simple|medium|complex)
- \/help

## Fixes from review
- Make \/help ack render real newlines (no literal \n).
- \/context no longer writes into `last_error`; it clears `reason`/`last_error` and resets attempts.
- \/priority clears `last_error` on success.
- Add an owner-feedback lock to avoid double-processing when poll loops overlap.

## Testing

Closes #151

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -1092,7 +1092,7 @@ process_owner_comment() {
       fi
       if [ -z "$ctx" ]; then
         db_task_update "$task_id" "$advance_ts"
-        _owner_cmd_ack "❌ Missing text for `/context`. Usage: `/context <text>` (or put text on following lines)."
+        _owner_cmd_ack "❌ Missing text for \`/context\`. Usage: \`/context <text>\` (or put text on following lines)."
       else
         append_task_context "$task_id" "### Owner context (${login:-owner} ${created_at})"$'\n'"${ctx}"$'\n---\n'
         db_task_update "$task_id" \
@@ -1140,7 +1140,7 @@ process_owner_comment() {
       ;;
     *)
       db_task_update "$task_id" "$advance_ts"
-      _owner_cmd_ack "❌ Unknown command: `/${cmd}`. Use `/help` for supported commands."
+      _owner_cmd_ack "❌ Unknown command: \`/${cmd}\`. Use \`/help\` for supported commands."
       ;;
   esac
 }


gabrielkoerich:

The implementation is functionally correct and covers all described slash commands with good test coverage for the happy paths. A few issues worth noting:

**1. poll.sh (lines ~97–129): Missing `new` and `routed` from candidate scan**

The poll loop scans only `done`, `in_review`, `needs_review`, `blocked`:
```bash
db_task_ids_by_status "done" || true
db_task_ids_by_status "in_review" || true
db_task_ids_by_status "needs_review" || true
db_task_ids_by_status "blocked" || true
```
Commands like `/assign`, `/priority`, and `/close` are also meaningful on `new` and `routed` tasks (e.g., reassigning an agent before a task runs, or closing a queued task). If this exclusion is intentional ("rescue-only" semantics), it should be documented; otherwise these two statuses should be added.

**2. lib.sh `process_owner_comment` (context handler ~line 1012): Inline args shadow trailing lines**

```bash
ctx=$(printf '%s' "$args" | sed 's/^[[:space:]]*//;s/[[:space:]]*$//')
if [ -z "$ctx" ]; then
    ctx=$(_trailing_context)
fi
```

If the user writes `/context brief note\nmore detail here`, only `brief note` is used and the trailing lines are silently dropped. The two sources should probably be combined (`ctx="${args}\n${trailing}"` after trimming) or the behaviour should be documented explicitly.

**3. lib.sh `process_owner_comment`: Inner function namespace pollution**

`_owner_cmd_ack` and `_trailing_context` are defined inside `process_owner_comment` but become global bash functions on each call, relying on dynamic scoping of local variables (`$repo`, `$task_id`, `$body`) to work correctly. This is a subtle bash-ism: it works because `_owner_cmd_ack` is always called within the same call stack as `process_owner_comment`, but it leaks names into global function space and could be confusing during debugging. Moving them to top-level helpers (with explicit parameters) would be safer.

**4. tests/orchestrator.bats: Missing error-path coverage**

No tests for:
- `/assign nonexistent` → expect error ack, no state change beyond advancing `gh_last_feedback_at`
- `/priority bogus` → same
- `/context` with no inline text and no trailing lines → expect error ack
- unknown command (e.g., `/blah`) → expect "Unknown command" ack

These aren't blocking (happy paths are well covered), but the error paths include non-trivial logic (e.g., `db_task_update` still called to advance the timestamp) that could silently regress.

**5. lib.sh `_parse_owner_slash_command` (minor): two-pass stdout parsing**

The function outputs `cmd` and `args` as two lines; the caller re-parses them with `sed -n '1p'` / `sed -n '2p'`. This works because `args` is derived from the first line of the body and can never contain a literal newline, but it's fragile. A named-pipe or `read` approach (e.g., `IFS=$'\n' read -r cmd args <<< "$parsed"`) would be cleaner.

**No blocking issues were found.** The core slash-command dispatch, lock-based deduplication, ack filtering (`<!-- orch: -->` marker), and sequential comment processing are all correctly implemented. The `gh_mock.sh` fix to include `user: {login: $login}` correctly aligns the mock with the jq projection in `fetch_owner_feedback`. The `$'...'` ANSI-C quoting for the `/help` response correctly produces real newlines.

---

gabrielkoerich:

The implementation is functionally correct and the slash-command dispatch logic is sound. A few issues worth noting:

**Bug: bare `/` treated as a non-command comment instead of being ignored**
`scripts/lib.sh` – `_parse_owner_slash_command`: when the body is just `/` (or `/ ` with trailing space), `[[ "$trimmed" == /* ]]` matches, the function returns 0 (success), but `cmd` ends up as an empty string. The caller's `if [ -z "$cmd" ]` guard then falls through to `process_owner_feedback`, so a bare slash triggers regular-feedback processing. The function should add `[ -n "$cmd" ] || return 1` before `printf '%s\n' "$cmd"` to reject this case.

**Missing tests for error/rejection paths**
`tests/orchestrator.bats` covers all seven happy paths and case-insensitivity, but there are no tests for:
- `/assign <invalid>` (e.g. `/assign gpt4`) — the error-ack branch
- `/priority <invalid>` — the error-ack branch
- `/context` with neither inline text nor trailing lines — the error-ack branch
- An unknown command like `/foo` — the catch-all `*` branch

These branches post error-ack comments and advance the timestamp, so verifying them matters.

**`LOCK_PATH` must be defined before calling `process_owner_feedback_for_task`**
`scripts/lib.sh` – `process_owner_feedback_for_task`: `fb_lock="${LOCK_PATH}.owner_feedback.${task_id}"`. If `LOCK_PATH` is empty, `mkdir` will create `.owner_feedback.<id>` in the working directory rather than failing loudly. An `[ -n "$LOCK_PATH" ] || return 1` guard (or a fallback default) would make this safer.

**Inner-function namespace leakage**
`scripts/lib.sh` – `process_owner_comment`: `_owner_cmd_ack` and `_trailing_context` are defined as named functions inside `process_owner_comment`. Bash named functions are global, so after the first invocation these names are permanently defined in the script's function table. They rely on dynamic scoping to access `repo`, `task_id`, and `body` from the outer function's locals — which works correctly here because they're only ever called within `process_owner_comment`, but it is fragile. If either helper is ever called outside that context, it will silently use whatever `repo`/`task_id` happen to be in scope. Worth converting them to local helper commands or passing arguments explicitly.

**Candidates list omits `routed` tasks in `scripts/poll.sh`**
The new polling block (poll.sh, approx. lines 100–128) only scans `done`, `in_review`, `needs_review`, and `blocked`. If an owner posts `/assign opencode` on a task that is already `routed` to a different agent, the command won't be processed until the task transitions to one of the listed states. This may be intentional (to avoid interrupting in-flight work), but it is not documented in the function comment or the docs table.

**Documentation note**
`docs/content/github-sync.md`: the `/priority` table entry shows `<low|medium|high>` but the implementation also accepts `simple`, `medium`, `complex` directly (via the fallthrough in the case statement). The doc could mention this for completeness.

None of these are blocking. The core dispatch, locking, timestamp-advancement, and ack-exclusion logic are all correct, and the test suite gives solid coverage of the primary paths.

---

PR #191, opened 2026-02-22

### LINKED ISSUE #153: Auto-reroute tasks to fallback agent on usage limit errors

gabrielkoerich:
## Summary

- Adds `is_usage_limit_error()` helper in `scripts/lib.sh` to detect rate/usage limit errors via regex (429, too many requests, rate-limit, quota exceeded, overloaded, etc.)
- Adds `pick_fallback_agent()` helper that rotates to the next available agent, skipping a configurable exclude list to prevent ping-pong
- Adds `reroute_on_usage_limit()` in `scripts/run_task.sh` that auto-reroutes tasks when usage/rate limits are detected:
  - On non-zero exit code with usage-limit output
  - On invalid/empty JSON response (likely agent crash due to limit)
  - On agent response with missing status field
  - On agent-reported needs_review/blocked with rate-limit reason
- Tracks `limit_reroute_chain` per-task to avoid bouncing between agents; chain clears on `done` or manual retry
- Posts GitHub issue comment with redacted snippet when rerouting or when no fallback is available
- Extracts auth/billing error path to use `pick_fallback_agent()` for consistency
- Adds `redact_snippet()` helper to strip API keys/tokens before posting to GitHub
- Clears `limit_reroute_chain` in `retry_task.sh` when task is manually reset to new

## Changes from PR review feedback (PR #188)
- Tightened `is_usage_limit_error` regex: removed overly broad `service unavailable` and `temporarily unavailable` patterns; kept only Anthropic-specific `service overloaded`
- Restored bare `\bquota\b` match that was accidentally dropped
- Added `redact_snippet()` to prevent accidental credential leaks in GitHub comments
- Fixed opencode free-model index in `reroute_on_usage_limit`: count opencode occurrences in chain instead of total chain length
- Restored history-based `FREE_IDX` in auth/billing path
- Clear `limit_reroute_chain` in `retry_task.sh` on manual requeue
- Added positive and negative BATS tests for `is_usage_limit_error`

## Test plan
- [x] 239 bats tests passing including 4 reroute-specific tests + 7 is_usage_limit_error unit tests
- [x] Auto-reroutes on usage limit (agent-reported needs_review)
- [x] Auto-reroutes on non-JSON response (crash/limit)
- [x] Preserves limit_reroute_chain on in_progress completions (ping-pong prevention)
- [x] 503 Service Unavailable does NOT trigger reroute (negative test)
- [x] Bare 'temporarily unavailable' does NOT trigger reroute (negative test)
- [x] Generic auth errors do NOT trigger reroute (negative test)

Closes #153

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -348,7 +348,19 @@ available_agents() {
 is_usage_limit_error() {
   local text="${1:-}"
   [ -n "$text" ] || return 1
-  printf '%s' "$text" | rg -qi '(?:\b429\b|too many requests|rate[ _-]?limit|usage[ _-]?limit|quota exceeded|insufficient[_ -]?quota|exceeded[_ -]?quota|limit (?:reached|exceeded)|overloaded[_ -]?error|service (?:overloaded|unavailable)|temporarily unavailable)'
+  # Match AI provider rate/quota errors. Patterns are intentionally specific to
+  # avoid false positives from generic network errors (e.g. 503 Service Unavailable,
+  # SSH timeouts). "service overloaded" is Anthropic-specific. "service unavailable"
+  # and bare "temporarily unavailable" are omitted — too common in unrelated errors.
+  printf '%s' "$text" | rg -qi '(?:\b429\b|too many requests|rate[ _-]?limit|usage[ _-]?limit|\bquota\b|insufficient[_ -]?quota|exceeded[_ -]?quota|limit (?:reached|exceeded)|overloaded[_ -]?error|service overloaded)'
+}
+
+# Redact common API key/token patterns before publishing text to GitHub comments.
+redact_snippet() {
+  printf '%s' "${1:-}" \
+    | sed -E 's/(sk|pk)-[A-Za-z0-9_-]{10,}/[REDACTED]/g' \
+    | sed -E 's/Bearer [A-Za-z0-9._~+/=-]{8,}/Bearer [REDACTED]/g' \
+    | sed -E 's/token=[A-Za-z0-9._~+%=-]{8,}/token=[REDACTED]/g'
 }
 
 # Pick a fallback agent from the locally available agents, rotating after the
--- a/scripts/retry_task.sh
+++ b/scripts/retry_task.sh
@@ -19,6 +19,6 @@ if [ "$STATUS" = "new" ] || [ "$STATUS" = "routed" ] || [ "$STATUS" = "in_progre
   exit 0
 fi
 
-db_task_update "$TASK_ID" "status=new" "agent=NULL"
+db_task_update "$TASK_ID" "status=new" "agent=NULL" "limit_reroute_chain=NULL"
 append_history "$TASK_ID" "new" "retried from $STATUS"
 log "[retry] task=$TASK_ID reset from $STATUS to new"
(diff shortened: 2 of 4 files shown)


gabrielkoerich:

The core feature (rerouting, ping-pong prevention, chain tracking) is correct and well-tested. Issues found are minor or edge-case. Specific notes below.

**`opencode_idx` logic is dead code (scripts/run_task.sh:816-828)**
The PR description says "Fixed opencode free-model index... count opencode occurrences in chain instead of total chain length." However, the fix doesn't achieve cycling in practice. When `next_agent == "opencode"`, opencode is guaranteed NOT to be in `chain` (because `pick_fallback_agent` excludes all entries in `chain`). So `opencode_idx` is always 0, and `next_model` is always `free_models[0]`. The cycling through free models never activates within a single reroute session. This is not a correctness bug — tasks still get rerouted correctly to opencode with model[0] — but the intended round-robin behaviour across free models does not work as described. Low impact but worth a follow-up.

**`limit (?:reached|exceeded)` pattern is overly broad (scripts/lib.sh:354)**
This alternation can match unrelated messages like "time limit reached", "memory limit exceeded", or "connection limit reached". Consider tightening to `(?:rate|usage|api)[_ -]?limit (?:reached|exceeded)` or similar to reduce false positives that could incorrectly trigger a reroute instead of surfacing the real error.

**`redact_snippet` missing common token formats (scripts/lib.sh:359-363)**
The three patterns cover `sk/pk-` prefixed keys, Bearer tokens, and `token=` URL params, but miss several common formats: Anthropic `sk-ant-api03-*` keys (the prefix doesn't start with `sk-` after the infix), GitHub PATs (`ghp_`, `ghs_`, `github_pat_`), AWS access keys (`AKIA*`). This is a best-effort function and the existing patterns catch the most likely leaks, but it is worth expanding incrementally. Low priority.

**Missing database schema migration (scripts/**)
The diff introduces `limit_reroute_chain` as a new task field but no `ALTER TABLE` migration is visible. The 239 passing BATS tests suggest the test harness creates a fresh DB that already includes this column (or the storage layer handles dynamic fields). For existing SQLite-backed deployments this column must be added. If the project uses a flexible key-value or JSON sidecar store the concern doesn't apply. Worth confirming in the PR description or adding a migration guard in the DB init path.

**No explicit test for 'no fallback available' outcome (tests/orchestrator.bats)**
The ping-pong test (@test line ~863) implicitly covers this path: after both agents hit their limits the task lands in `needs_review`. But an explicit assertion that status == `needs_review` and `limit_reroute_chain` contains all tried agents when no fallback exists would make the contract clearer and prevent regression.

**Minor: `current` skip in `pick_fallback_agent` is redundant when chain is non-empty (scripts/lib.sh:387-389)**
`current` is always included in `chain` before `pick_fallback_agent` is called (from `reroute_on_usage_limit`), so the `exclude_csv` check already excludes it. The explicit `current` equality check is still correct and needed for the auth/billing call site where `exclude_csv` is empty, so this is a nit only.

All four reroute-specific integration tests and seven unit tests for `is_usage_limit_error` are correct and cover the important paths (happy path, non-JSON crash, ping-pong prevention, chain preservation across `in_progress`). The retry chain clear in `retry_task.sh` is correct. Overall the implementation is sound.

---

PR #199, opened 2026-02-22
fix(status): align totals with status counts

### LINKED ISSUE #196: Respond to @orchestrator mention in #150
This task was created from an @orchestrator mention.

- **Repo:** `gabrielkoerich/orchestrator`
- **Target:** #150
- **Author:** @gabrielkoerich
- **Comment:** https://github.com/gabrielkoerich/orchestrator/issues/150#issuecomment-3939530591
- **Created:** 2026-02-21T21:25:55Z

### Mention Body

```markdown
Implemented @orchestrator mention-triggered tasks + response mirroring.

- Added `scripts/gh_mentions.sh` to poll repo-wide issue/PR comments since last run, detect actionable `@orchestrator` mentions (skips fenced code + blockquotes + orchestrator-marked comments), and create a new task issue with captured context.
- Added idempotency via a persisted comment-id → task-id map in the project state dir to prevent duplicate tasks from the same mention.
- Updated GitHub backend to mirror the agent’s structured result comment back onto the originating issue/PR thread (deduped separately).
- Wired the mention poller into `scripts/serve.sh`.
- Added bats coverage + extended `tests/gh_mock.sh` for the new repo-wide comments endpoint.

PR: #178
Worktree: /Users/gb/Projects/orchestrator/.orchestrator/worktrees/gh-task-150-orchestrator-mentions-in-github-comments
```

### Instructions

Respond back on #150 with your results and next steps.

gabrielkoerich:
## Summary

- fix(status): align totals with status counts
- fix(mentions): prevent duplicate tasks across projects

## Testing

Closes #196

### DIFF
--- a/scripts/status.sh
+++ b/scripts/status.sh
@@ -22,14 +22,9 @@ if [ "$IS_JSON" = true ]; then
 fi
 
 if [ "$IS_GLOBAL" = true ]; then
-  TOTAL=$(db_total_task_count)
+  : # counts below are global by default in the GitHub backend
 else
-  TOTAL=$(db_total_filtered_count)
-fi
-
-if [ "$TOTAL" -eq 0 ]; then
-  echo "No tasks. Add one with: orchestrator task add \"title\""
-  exit 0
+  : # counts below are filtered by PROJECT_DIR in non-GitHub backends
 fi
 
 _count() {
@@ -44,6 +39,12 @@ BLOCKED=$(_count "blocked")
 DONE=$(_count "done")
 NEEDS_REVIEW=$(_count "needs_review")
 OPEN_TOTAL=$((NEW + ROUTED + INPROG + IN_REVIEW + BLOCKED + NEEDS_REVIEW))
+TOTAL=$((OPEN_TOTAL + DONE))
+
+if [ "$TOTAL" -eq 0 ]; then
+  echo "No tasks. Add one with: orchestrator task add \"title\""
+  exit 0
+fi
 
 {
   printf 'STATUS\tQTY\n'
--- a/tests/orchestrator.bats
+++ b/tests/orchestrator.bats
@@ -2831,6 +2831,46 @@ JSON
   [[ "$output" == *"total"* ]]
 }
 
+@test "status.sh shows correct per-status counts and totals" {
+  ROUTED_OUTPUT=$("${REPO_DIR}/scripts/add_task.sh" "Routed" "" "")
+  ROUTED_ID=$(_task_id "$ROUTED_OUTPUT")
+  tdb_set "$ROUTED_ID" status "routed"
+
+  IP1_OUTPUT=$("${REPO_DIR}/scripts/add_task.sh" "In Progress 1" "" "")
+  IP1_ID=$(_task_id "$IP1_OUTPUT")
+  tdb_set "$IP1_ID" status "in_progress"
+
+  IP2_OUTPUT=$("${REPO_DIR}/scripts/add_task.sh" "In Progress 2" "" "")
+  IP2_ID=$(_task_id "$IP2_OUTPUT")
+  tdb_set "$IP2_ID" status "in_progress"
+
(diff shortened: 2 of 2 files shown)


gabrielkoerich:

## Review of PR #199

### `scripts/status.sh`

**Fix is correct.** The original bug: `db_total_task_count` / `db_total_filtered_count` could return a number inconsistent with what `_count()` returns (different query paths, caching, filtering). Computing `TOTAL=$((OPEN_TOTAL + DONE))` from the same `_count()` calls used for the table guarantees the displayed total always equals the sum of per-status rows.

The no-op `: # counts below are ...` comments (lines ~25-29) replacing the old `if/else` block are valid bash. The `IS_GLOBAL` flag is presumably still consumed by `_count()` internally, which is the right place for that distinction now.

One minor note: the `if [ "$TOTAL" -eq 0 ]` early-exit guard is now placed *after* all `_count()` calls (~line 42). This means all DB queries run even when there are no tasks. For a CLI tool this overhead is negligible, and the correctness gain is worth it.

---

### `scripts/gh_mentions.sh`

**Root cause fix is correct.** The original bug: all project configs sharing the same `STATE_DIR` wrote to the same `mentions.json`, so the second project's poll would find the comment already in `processed{}` and skip creating its task. The fix gives each repo its own DB file at `${ORCH_HOME}/.orchestrator/mentions/${repo_key}.json`, shared across all projects polling that repo.

**Specific observations:**

1. **`MENTIONS_DB_PATH` / `MENTIONS_LOCK_DIR` scoping (lines ~131-135):** These are set without `local` inside `main()`, which is intentional — they need to be visible to the helper functions. Since this is a script entry point called once, the global-namespace leak is harmless.

2. **`tr '/:' '__'` repo_key generation (line ~132):** Both `/` and `:` map to `_`, so `owner/repo` becomes `owner_repo`. GitHub repo identifiers cannot contain `:` or `/`, so no collision is possible between valid repo names. Sound.

3. **`updated_at` for `max_seen` (lines ~166-169):** Using `updated_at` (falling back to `created_at`) for advancing the `since` cursor is a behavior change. If an old comment is edited, the cursor could advance backward conceptually — however the dedup check via `mentions_db_lookup_task` prevents re-processing, so correctness is maintained. The tradeoff is that more comments may be re-fetched from the API after an edit. This is acceptable.

4. **Guard clauses (`[ -n "$path" ] || return 1`):** All helper functions guard against an unset `MENTIONS_DB_PATH`. Since `main()` always sets it before calling any helper, these guards are defensive-only and correct.

5. **`acquire_mentions_lock` `dirname` call (line ~73):** `dirname` of `${mentions_dir}/${repo_key}.lock` correctly resolves to `${mentions_dir}`, which is the right parent to `mkdir -p`. Sound.

6. **Potential gap:** If `ORCH_HOME` is unset, `mentions_dir` becomes `/.orchestrator/mentions`. This is pre-existing behavior (not introduced by this PR) and presumably `ORCH_HOME` is always initialized by the orchestrator bootstrap.

---

### `tests/orchestrator.bats`

**Idempotency test (line ~2030):** Correctly exercises two distinct `PROJECT_DIR` values polling the same mock repo, then asserts `tdb_count` returns 2 (init task + one mention task, not two mention tasks). Solid.

**Status counts test (line ~2831):** Creates one task per status variant, sets them explicitly, then uses `grep -qE` to assert each row. Using `grep -qE` without `run` means a non-matching pattern directly fails the bats test — this is the correct pattern for inline assertions. The regex `'^new[[:space:]]+1[[:space:]]*$'` correctly matches tab-separated output from `printf 'STATUS\tQTY\n'`. The expected counts (new=1, routed=1, in_progress=2, in_review=1, blocked=0, needs_review=1, done=1, open=6, total=7) are arithmetically consistent with the task setup.

---

**Summary:** Both fixes are logically sound, the tests cover the specific regressions being fixed, and no security issues or edge-case bugs were identified. Approve.

---

PR #224, opened 2026-02-22
fix(mentions): ignore mention-task cascades

### LINKED ISSUE #220: Respond to @orchestrator mention in #205
This task was created from an @orchestrator mention.

- **Repo:** `gabrielkoerich/orchestrator`
- **Target:** #205
- **Author:** @gabrielkoerich
- **Comment:** https://github.com/gabrielkoerich/orchestrator/issues/205#issuecomment-3940205208
- **Created:** 2026-02-22T05:22:45Z

### Mention Body

```markdown
Starting investigation of @orchestrator mention on PR #183. Worktree: /Users/gb/Projects/orchestrator/.orchestrator/worktrees/gh-task-205-respond-to-orchestrator-mention-in-183
```

### Instructions

Respond back on #205 with your results and next steps.

gabrielkoerich:
## Summary

- fix(mentions): ignore mention-task cascades

## Testing

Closes #220

### DIFF
--- a/scripts/gh_mentions.sh
+++ b/scripts/gh_mentions.sh
@@ -70,6 +70,23 @@ acquire_mentions_lock() {
   return 1
 }
 
+is_mention_task_issue() {
+  local repo="$1" issue_number="$2"
+  # Avoid mention-task cascades: comments on a "Respond to @orchestrator mention..." issue should not spawn more tasks.
+  local issue_json title body
+  issue_json=$(gh_api -X GET "repos/${repo}/issues/${issue_number}" 2>/dev/null || echo "")
+  [ -n "$issue_json" ] || return 1
+  title=$(printf '%s' "$issue_json" | jq -r '.title // ""' 2>/dev/null || echo "")
+  body=$(printf '%s' "$issue_json" | jq -r '.body // ""' 2>/dev/null || echo "")
+  if printf '%s' "$title" | rg -q '^Respond to @orchestrator mention in #[0-9]+'; then
+    return 0
+  fi
+  if printf '%s' "$body" | rg -q 'This task was created from an @orchestrator mention\\.'; then
+    return 0
+  fi
+  return 1
+}
+
 mention_actionable() {
   local body="${1:-}"
   [ -n "$body" ] || return 1
@@ -181,6 +198,10 @@ main() {
     local existing
     existing=$(mentions_db_lookup_task "$comment_id")
     if [ -n "$existing" ]; then
+      continue
+    fi
+
+    if is_mention_task_issue "$repo" "$issue_number"; then
       continue
     fi
 
--- a/tests/orchestrator.bats
+++ b/tests/orchestrator.bats
@@ -2075,6 +2075,32 @@ YAML
   [ "$output" -eq 2 ]
 }
 
+@test "@orchestrator mention on a mention-task issue is ignored" {
+  # Seed: create a mention task issue
+  run gh api "repos/mock/repo/issues/${INIT_TASK_ID}/comments" -f body="hey @orchestrator please take a look"
+  [ "$status" -eq 0 ]
+  run gh_mentions.sh
+  [ "$status" -eq 0 ]
+
+  MENTION_TASK_ID=$(_task_id_by_title "Respond to @orchestrator mention in #${INIT_TASK_ID}")
+  [ -n "$MENTION_TASK_ID" ]
+
+  # Meta mention inside the mention task issue should not spawn another task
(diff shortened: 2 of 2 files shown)


gabrielkoerich:

The implementation is correct and addresses the cascade problem cleanly.

**`scripts/gh_mentions.sh` — `is_mention_task_issue` (lines 73–91 in diff)**

- Both the title pattern (`^Respond to @orchestrator mention in #[0-9]+`) and the body sentinel (`This task was created from an @orchestrator mention\.`) are reliable signals since they are produced by the orchestrator itself, not by external users.
- Error handling is safe: if `gh_api` fails, `issue_json` is empty, and the function returns 1 ("not a mention-task issue"). This means the cascade guard silently degrades — a spurious task might be created — rather than silently dropping real mentions. Fail-open is the correct default here.
- The `printf '%s' "$title" | rg -q '...'` pattern is safe from injection; user-controlled content is never interpolated into shell commands.

**Placement in the main loop (lines 201–204 in diff)**

The check is inserted AFTER `mentions_db_lookup_task` but BEFORE task creation. One thing worth verifying: `max_seen` is updated at line 161–162 (before ANY of the skip checks), so even when `is_mention_task_issue` causes a `continue`, the since-cursor still advances past that comment's `created_at`. Confirmed by reading `scripts/gh_mentions.sh:161-162` in the base file. This means the ignored comment will not be re-fetched on subsequent runs and the extra `gh_api` call is bounded to a single run per comment — not a recurring cost.

**`tests/orchestrator.bats` (lines 2078–2106 in diff)**

- The test correctly seeds the two-step scenario: real mention → mention-task created → meta-mention on mention-task → no second task.
- `tdb_count` assertion `[ "$output" -eq 2 ]` correctly reflects the expected state (1 original task + 1 mention task, no third task).
- The explicit `_task_id_by_title "Respond to @orchestrator mention in #${MENTION_TASK_ID}"` / `[ -z "$output" ]` assertion is a good belt-and-suspenders check that no cascade task was created.

**Minor observation (non-blocking):** The two detection conditions (title vs. body) provide independent redundancy, which is fine. However, in practice only the title check is needed since the title is always set by the orchestrator. The body check is a useful fallback if the title were ever changed, so keeping both is reasonable.

No correctness bugs, no security issues, test coverage is solid.

---

gabrielkoerich:

The PR correctly addresses the mention-task cascade problem. Here is a detailed review:

**`scripts/gh_mentions.sh` — `is_mention_task_issue()` (lines ~73–92)**

1. **Correctness**: The title pattern `'^Respond to @orchestrator mention in #[0-9]+'` in `rg` exactly matches the title template used in `main()` at line ~207: `title="Respond to @orchestrator mention in #${issue_number}"`. The two patterns stay in sync, which is the right approach.

2. **Safe failure default**: If `gh_api` returns nothing (line ~76: `|| echo ""`), the guard `[ -n "$issue_json" ] || return 1` causes the function to return 1 (not a mention-task), so the mention is still processed. This conservative default is correct — an API failure should not silently suppress mentions.

3. **Minor redundancy — body check**: The body pattern check (lines ~85–88) is a useful belt-and-suspenders fallback if the title format ever changes, but it introduces a second `printf | rg` invocation on every call. It's not wrong; just worth noting it adds minor overhead.

4. **Performance concern (non-blocking)**: `is_mention_task_issue` issues a full `gh_api` round-trip for *every* mention processed (line ~76). If there is a burst of mentions across many issues, this doubles the API calls and could approach secondary rate limits. A lightweight mitigation would be to cache issue metadata already fetched earlier in the loop, or to skip the lookup when the mention itself already carries enough context to identify the source. This is not a blocker for the current change, but worth a follow-up if rate limiting becomes an issue.

5. **`rg` dependency**: The function introduces a hard dependency on `rg` (ripgrep) for two pattern checks. Since `rg` is already used elsewhere in the script, this is consistent, but it should be documented as a runtime requirement if it isn't already.

**`scripts/gh_mentions.sh` — `main()` (lines ~201–204)**

The placement of the `is_mention_task_issue` guard is correct — it fires *before* any task creation logic, so no partial work is done for a cascade mention.

**`tests/orchestrator.bats` — new test (lines ~2075–2106)**

The test is well-structured:
- Creates a legitimate mention → verifies `MENTION_TASK_ID` is created.
- Posts a *meta-mention* inside the mention-task issue.
- Re-runs `gh_mentions.sh` and asserts `tdb_count` is still 2 (not 3).
- Explicitly asserts no task titled `"Respond to @orchestrator mention in #${MENTION_TASK_ID}"` was created.

This directly exercises the code path added in `main()`. Both positive (task created for the first mention) and negative (no task for the cascade) outcomes are checked.

**Summary**: The logic is sound, error handling is safe, and the test covers the critical path. The extra API call per mention is a minor concern worth monitoring but does not warrant blocking the PR.

---

PR #308, opened 2026-02-22
fix(mentions): dedup skips re-trigger on closed issues even when prior tasks are done

### LINKED ISSUE #306: fix(mentions): dedup should check if existing task is done AND target issue is closed before skipping
The per-issue dedup check in gh_mentions.sh looks for active mention tasks in the database. But there are two gaps:

## Gap 1: Tasks stuck in in_progress are treated as "active"

If a task returned `in_progress` for a mention on a closed issue, the dedup check sees it as "active" and correctly skips new tasks for that issue. This is correct behavior.

But if the task is marked `done`, new @orchestrator mentions on the SAME closed issue could re-trigger a task (since the old task is `done` the dedup allows it).

## Gap 2: The closed-issue check happens per-comment, not per-task

The `skip comments on closed issues` fix iterates over comments and skips if the issue is closed. But the dedup database check (lines 229-244) allows new tasks if all prior tasks for that issue are `done`. If someone comments `@orchestrator` on a closed issue, multiple times, each new comment could create a new task.

## Fix

After the closed-issue check at line 202-204, also skip mentions that reference `done` tasks on closed issues:
```bash
# After creating task, note the issue was closed so future comments are also skipped
```

And in the dedup check, if the target issue is closed (checked via cache), skip regardless of prior task status.

This prevents the scenario where a closed issue with multiple @orchestrator mentions keeps creating tasks.

gabrielkoerich:
## Summary

- **Gap 1 (max_seen not advancing)**: The `max_seen` timestamp was not updated for closed-issue comments because the `continue` fired before the update. This caused those comments to be re-fetched on every poll cycle without ever creating tasks.
- **Gap 2 (stale cache bypass)**: When all prior mention tasks for an issue were `done`, a new comment on the same (now-closed) issue could create another task — especially if the initial closed-issue check used a stale 120s-cached `open` response.

## Changes

- Moved `max_seen` update before the closed-issue `continue` so closed-issue comments always advance `since`
- In the per-issue dedup block, if any prior tasks exist for an issue, re-check the issue state **without cache**. If the issue is now closed, skip creation regardless of prior task status

## Test plan

- [x] `@orchestrator mention on closed issue is skipped` — no task created for comment on closed issue
- [x] `@orchestrator mention on closed issue with done task does not re-trigger` — after prior task completes and issue is closed, new comment does not create another task
- [x] `@orchestrator mention on closed issue advances since so comments are not re-fetched` — `since` advances past closed-issue comments
- [x] All 268 existing tests still pass

Closes #306

### DIFF
--- a/scripts/gh_mentions.sh
+++ b/scripts/gh_mentions.sh
@@ -196,17 +196,17 @@ main() {
     issue_number=$(printf '%s' "$issue_url" | sed -E 's#.*/issues/([0-9]+).*#\1#')
     [ -n "$issue_number" ] || continue
 
+    local seen_at
+    seen_at="${updated_at:-$created_at}"
+    if [ -n "$seen_at" ] && [ "$seen_at" \> "$max_seen" ]; then
+      max_seen="$seen_at"
+    fi
+
     # Skip comments on closed issues — no point creating tasks for them
     local _issue_state
     _issue_state=$(gh_api "repos/${repo}/issues/${issue_number}" --cache 120s -q '.state' 2>/dev/null || true)
     if [ "$_issue_state" = "closed" ]; then
       continue
-    fi
-
-    local seen_at
-    seen_at="${updated_at:-$created_at}"
-    if [ -n "$seen_at" ] && [ "$seen_at" \> "$max_seen" ]; then
-      max_seen="$seen_at"
     fi
 
     if ! printf '%s' "$comment_body" | rg -qi "@orchestrator"; then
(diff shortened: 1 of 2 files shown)


gabrielkoerich:

The PR correctly addresses both described gaps. Here is the detailed analysis:

**Gap 1 Fix — `max_seen` before `continue` (scripts/gh_mentions.sh ~line 196-210)**

The original code updated `max_seen` after the closed-issue `continue`, which meant closed-issue comments never advanced `since`. The fix is correct: move the `seen_at` block to before the `if [ "$_issue_state" = "closed" ]` guard. This is a clean, minimal change with no side effects.

**Gap 2 Fix — Fresh state re-check when prior tasks exist (scripts/gh_mentions.sh ~line 230-244)**

The new block fires when `_active_tasks` is non-empty (any prior tasks for the issue, regardless of status), then calls the GitHub API without `--cache 120s` to get a fresh issue state. If closed, it skips task creation. The logic is sound:
- Safe failure mode: if `_fresh_state` is empty (API failure), the condition `[ "$_fresh_state" = "closed" ]` is false and task creation proceeds — correct fail-open behavior.
- The check is intentionally broader than strictly needed (it fires for done *and* active prior tasks). This is actually safer — you wouldn't want a new task on a closed issue even if an existing task is still running.

**Minor observation (non-blocking):** The new skip path uses `log_err` (`echo ... >&2`) to signal a successful/expected skip. The existing code in line 242 already uses `log_err` for the analogous 'active task exists' skip, so this is consistent with the existing style. Neither case is a real error, but changing it would be scope creep.

**Tests (tests/orchestrator.bats ~line 2327-2418)**

Three new tests are added:
1. `@orchestrator mention on closed issue is skipped` — Validates the basic closed-issue skip path.
2. `@orchestrator mention on closed issue with done task does not re-trigger` — Directly tests Gap 2: prior task goes to `done`, issue is closed, new comment arrives, no new task created. This is the critical regression test.
3. `@orchestrator mention on closed issue advances since so comments are not re-fetched` — Validates `since` advances after the first poll and stays stable on the second poll, confirming Gap 1 is fixed. Assertion `[ "$since_before" = "$since_after" ]` verifies idempotency; `[ "$since_after" != "1970-01-01T00:00:00Z" ]` confirms `since` actually advanced. Coverage is sufficient.

All 268 existing tests still pass per the PR description. No correctness bugs, security issues, or missing edge cases found. Approve.

---

PR #310, opened 2026-02-22
Rust core

gabrielkoerich:


gabrielkoerich:

This is a substantial PR introducing a Rust core. The architecture is sound and the PLAN.md is well-structured, but there are several correctness bugs and important gaps to address before merging.

**Critical Bugs**

1. **Deadlock in `src/engine/runner.rs` lines ~120-160**: After `child.wait()` returns, the code attempts to read stdout/stderr from the child. If `run_task.sh` produces enough output to fill the OS pipe buffer (typically 64 KB), `child.wait()` will never return because the subprocess is blocked trying to write. This is a well-known subprocess deadlock. Fix: use `child.wait_with_output()` to drain stdout/stderr concurrently with waiting, or spawn tasks to consume the pipes before calling `wait()`.

2. **Race condition in `src/backends/github.rs` lines ~50-62 (`update_status`)**: The implementation removes the old `status:*` label(s) and then adds the new one in two separate API calls. There is a window between these calls where the issue has no status label. During that window, another poll tick could see the issue as unlabelled and re-dispatch it. Worse, if `add_labels` fails after `remove_label` succeeds, the task permanently loses its status label. Fix: use a single `gh api` call that atomically sets the labels array, or tolerate partial failure by re-fetching current labels and retrying.

**Truncated Diff — Incomplete Review**

The diff is cut off mid-function in `src/github/cli.rs` (the `create_issue` function body is incomplete). Files referenced from multiple modules are entirely absent from the diff: `src/main.rs`, `src/tmux.rs` (referenced in `engine/mod.rs`), `src/security.rs` (whitelisted in `.gitleaks.toml`), `src/parser.rs`, and `src/sidecar.rs`. A complete security and correctness review is not possible without these files. The `TmuxManager` API (`snapshot()`, `session_name()`, `kill_session()`, `list_sessions()`, `session_exists()`) is called throughout `engine/mod.rs` but its implementation is not reviewable.

**CI Regressions**

3. **`.github/workflows/release.yml` line 6**: The `pull_request` trigger is removed. CI will no longer run automatically on PRs, which means the `cargo test` and `cargo clippy` gates only run on push to the default branch. This is a significant regression for code review quality.

4. **`.github/workflows/release.yml`**: The `release` and `deploy` jobs are deleted with no replacements added in this PR. Auto-versioning, Homebrew formula updates, and GitHub Pages deployment are now entirely absent. If this is intentional (to be added in a follow-up), it should be clearly noted and tracked.

**Medium Issues**

5. **`src/engine/runner.rs` lines ~70-95 (`TaskRunner::new`)**: The scripts-discovery logic silently falls back through multiple paths ending in `current_dir()`. If the fallback is used unintentionally, tasks will silently fail with a misleading 'run_task.sh not found' error. The fallback chain should emit a warning at each stage, and the final fallback should be more explicit.

6. **`src/engine/mod.rs` tick function (~line 180-190)**: `tmux.snapshot()` returns `(task_id, active)` pairs and kills sessions where `active == false`. Without seeing `TmuxManager::snapshot()`, it is unclear how `active` is determined and whether a newly-dispatched task (whose session was just spawned) could be incorrectly classified as inactive and killed immediately.

7. **`src/engine/jobs.rs` lines ~155-180 (`bash` job type)**: `tokio::process::Command::new("bash").arg("-c").arg(cmd)` runs arbitrary shell commands from `jobs.yml`. This is intentional per the design, but deserves a comment noting the trust model (only operator-controlled YAML).

**Minor Issues**

8. **`src/channels/mod.rs` lines ~105-118**: `ChannelRegistry` does not implement `Default` despite only containing a `Vec`. Since `ChannelRegistry::new()` is provided, add `impl Default for ChannelRegistry` delegating to `new()` for idiomatic Rust.

9. **`src/engine/runner.rs` line ~150**: After a timeout, `child.kill().await.ok()` silently ignores kill errors. These should at minimum be logged since a zombie child holding a semaphore permit would affect the parallelism limit.

---

gabrielkoerich:

## Critical

### `src/engine/runner.rs` — deadlock on large stdout/stderr

Lines 87–133: the code pipes both stdout and stderr (`Stdio::piped()`) but only reads them *after* `child.wait()` returns:

```rust
let mut child = Command::new("bash")
    .stdout(std::process::Stdio::piped())
    .stderr(std::process::Stdio::piped())
    .spawn()?;

let status = match timeout(Self::TASK_TIMEOUT, child.wait()).await { ... };

// reads happen AFTER wait — too late
let stdout = match child.stdout.take() { ... read_to_end ... };
let stderr = match child.stderr.take() { ... read_to_end ... };
```

If `run_task.sh` writes more than ~64 KB (Linux) or ~128 KB (macOS) to stdout or stderr — extremely likely for full agent runs with LLM responses and git output — the OS pipe buffer fills, the child process blocks on its write, and `child.wait()` blocks waiting for the child to exit. Classic deadlock; neither side proceeds.

Fix: use `child.wait_with_output()` which drains stdout/stderr concurrently while waiting:

```rust
let output = match timeout(Self::TASK_TIMEOUT, child.wait_with_output()).await {
    Ok(result) => result.context("waiting for run_task.sh")?,
    Err(_) => {
        // child.kill() is no longer available after wait_with_output takes ownership
        // spawn and kill before calling wait_with_output, or restructure
        anyhow::bail!("run_task.sh timed out after 30 minutes");
    }
};
if !output.status.success() { ... use output.stdout / output.stderr ... }
```

Alternatively, if capturing output isn't needed for correctness, use `Stdio::inherit()` for both stdout and stderr and avoid the issue entirely.

---

## Medium

### `.github/workflows/release.yml` — `pull_request` trigger removed

Line 5 of the original (`pull_request:`) was deleted. CI (fmt, clippy, tests) now only runs on direct `push` and manual `workflow_dispatch`. PRs to this repo will not receive automated checks, removing the primary guard against merging broken code. Add `pull_request:` back as a trigger.

### `.github/workflows/release.yml` — release and deploy jobs entirely removed

The full release automation (semver tagging, GitHub release creation, Homebrew tap update) and the GitHub Pages deploy were deleted. This may be intentional during migration, but it leaves no release path in CI. If this is intentional, it should be documented explicitly and a follow-up issue should track re-adding the release job for the Rust binary (with cross-compilation for arm64/x86_64 darwin).

---

## Low / Minor

### `src/github/cli.rs` — no pagination in `list_issues`

The `list_issues` implementation passes `per_page=100` with no cursor or page parameter. Repos with more than 100 open tasks in a given status will silently drop tasks. At minimum this should be documented; ideally it should loop with `page` increments or use `--paginate` via the `gh` CLI.

### `src/engine/mod.rs` — Phase 4 and `sync_tick` are stubs

Phase 4 (unblocking parent tasks, lines ~242–250) and `sync_tick` (lines ~265–278) are entirely `TODO` placeholders. The v0 equivalents (`cleanup_worktrees.sh`, `review_prs.sh`, `gh_mentions.sh`) are no longer in CI. While the PLAN.md describes these as future phases, the engine serving in this state means worktree cleanup and PR merge detection — which are needed for task lifecycle correctness — are silently skipped. This should be explicitly noted in the PR description and a tracking issue created.

### `src/db.rs:38` — WAL pragma on in-memory DB is a no-op

`open_memory()` calls `PRAGMA journal_mode=WAL` on an in-memory connection. SQLite silently ignores this pragma for `:memory:` databases. The statement is harmless but misleading; it should be removed or guarded with a `if path != ":memory:"` check.

### `src/engine/runner.rs:58–76` — hardcoded `/opt/homebrew` path

The Homebrew Cellar path is hardcoded to `/opt/homebrew/Cellar` (Apple Silicon). Intel Macs use `/usr/local/Cellar`. Use `brew --prefix` at runtime or check `$HOMEBREW_CELLAR` / `$HOMEBREW_PREFIX` env vars for portability.

### `src/engine/jobs.rs:114–131` — only `status:done` clears `active_task_id`

The check treats any status other than `status:done` (including `status:blocked`, `status:needs_review`, `status:in_review`) as "active", which prevents a cron job from firing again if its previous task is stuck in review indefinitely. Consider whether `status:blocked` and `status:needs_review` should also clear the lock, or add a configurable timeout.

### Diff truncation note

The diff was truncated at 5000 lines in the original submission; full content of `src/main.rs`, `src/github/cli.rs`, `src/tmux.rs`, `src/security.rs`, and `src/sidecar.rs` was retrieved separately. `src/tmux.rs` exists and exports `TmuxManager` as expected by `src/engine/mod.rs` — no missing symbol issues.

---

gabrielkoerich:

This is a well-architected PR with a clear migration plan, but it has several correctness issues, a security regression, and a blocking deployment bug that should be fixed before merge.

## Critical

### 1. Security regression: removal of shellcheck and semgrep CI jobs
`.github/workflows/release.yml` removes the `shellcheck` and `semgrep-and-secrets` jobs entirely. According to PLAN.md and the code in `src/engine/runner.rs`, bash scripts (`run_task.sh`, git ops, etc.) are explicitly kept in this PR. Removing static analysis for those scripts while they remain in the critical path is a net security regression. Consider keeping shellcheck at a minimum, or document explicitly that bash scripts have been moved out of scope.

### 2. `src/security.rs` added to gitleaks exclusion without explanation
`.gitleaks.toml` line adds `'''^src/security\.rs$'''` to the exclusion list. This file does not appear in the diff at all — we cannot see its contents. Adding a Rust source file to gitleaks exclusion is a significant flag and requires justification. If the file contains test fixtures, they should live in `tests/` (already excluded). If it contains real credentials or keys that are 'expected', that needs to be documented clearly in the PR.

### 3. Formula SHA256 is all zeros — blocks actual deployment
`Formula/orch.rb` line 5: `sha256 "0000000000000000000000000000000000000000000000000000000000000000"`. Homebrew will reject this on install. This should either be removed from the PR scope or filled in via the release CI step.

## Important

### 4. Hardcoded old package name in `src/engine/runner.rs:65`
```rust
let brew_path = PathBuf::from("/opt/homebrew/Cellar").join("orchestrator");
```
This PR renames the package to `orch` (see formula rename, CI formula path changes), but the runner still looks for the old `orchestrator` cellar path. This will silently fall back to `orch_home.join("scripts")`, which may not exist. Should be `orch` or preferably driven by a constant.

### 5. No in-memory caching in `src/config/mod.rs`
The `get()` function reads YAML from disk on every call. PLAN.md explicitly lists "In-memory struct, hot-reload" as a benefit of the Rust migration, and the performance table claims config reads drop from ~20ms to ~0.1ms. The current implementation does not achieve this — it will read and parse YAML files on every invocation. At minimum, this should be noted as a known limitation or a follow-up issue should be tracked.

### 6. No pagination in `src/github/cli.rs:list_issues` (line ~67)
```rust
"-f", "per_page=100",
```
Only the first page of 100 issues is fetched. If any status bucket (e.g., `status:new`) has more than 100 open issues, tasks will silently be missed. The GitHub Issues API supports pagination via `Link` headers. This is a correctness issue for projects with high issue volume.

### 7. No tests for `TaskRunner` in `src/engine/runner.rs`
This is the critical execution path — the code that actually spawns `run_task.sh` and monitors it. The file has zero tests. The `project_dir()` function also hardcodes `~/Projects/` (line ~109) as a fallback which will be wrong for most users. At minimum, unit tests for path resolution logic and a note on how to test the runner would be helpful.

### 8. `repo` fallback to `"owner/repo"` in `src/engine/mod.rs:78`
```rust
let repo = crate::config::get("repo").unwrap_or_else(|_| "owner/repo".to_string());
```
This silently swallows a missing config and proceeds with an invalid repo slug. Every subsequent GitHub API call will fail at runtime with a confusing error. This should either `?`-propagate the config error or print a clear startup message.

## Minor

### 9. `transport` is created but never used in `src/engine/mod.rs:97`
`let transport = Arc::new(Transport::new())` is assigned but only explicitly dropped at shutdown. It is never passed to any channel or used for message routing. This is clearly placeholder code but `cargo clippy` with `-D warnings` (as configured in CI) will flag the unused `Arc` clone. Consider adding a `// TODO: wire channels` comment or removing until channels are implemented.

### 10. `remove_label` swallows all errors in `src/github/cli.rs:~121`
The docstring says it returns `Ok` for 404 (label not found), but the implementation swallows ALL errors with a warning log. Errors like 401 (token expired), 403 (permissions), or 500 (server error) will be silently ignored, causing status updates to appear successful while leaving the issue in an inconsistent label state. Consider only ignoring 404.

### 11. Version mismatch: CI produces `v1.0.0` but `Cargo.toml` is `1.0.0-alpha.1`
The release CI change (line ~124) sets `NEW_VERSION="v1.0.0"` as the baseline when no prior tags exist. `Cargo.toml` declares `version = "1.0.0-alpha.1"`. These will diverge on first release. The CI versioning logic should either read from `Cargo.toml` or the alpha tag should be reflected in the CI default.

---

gabrielkoerich:

## PR Review: Rust core

**Note:** The diff is truncated at 5000 lines. `src/security.rs`, `src/parser.rs`, `src/tmux.rs`, and the remainder of `src/main.rs` are not visible. Several of the issues below are based on what *is* visible.

---

### Critical / High

**1. `src/security.rs` excluded from gitleaks without visible content**  
`.gitleaks.toml` adds `^src/security\.rs$` to the scan-ignore list, but `src/security.rs` is not included in the truncated diff. The only legitimate reason to exclude a *source* file from secret scanning is that it contains test fixture strings that look like secrets (e.g., regex patterns). If that is the case, add an inline comment explaining why. If it is not the case, this exclusion must be removed — it defeats the purpose of the scan for this file.

**2. `remove_label` always returns `Ok(())`, making `?` in `update_status` a no-op — perpetuating the label-sprawl bug**  
`src/github/cli.rs:119-127`:
```rust
pub async fn remove_label(...) -> anyhow::Result<()> {
    ...
    if let Err(e) = self.api(...).await {
        tracing::warn!(...);
    }
    Ok(())   // always succeeds
}
```
Then in `src/backends/github.rs:50-55`:
```rust
self.gh.remove_label(&self.repo, &id.0, label).await?;  // ? is dead
```
The `?` never propagates because `remove_label` eats every error. A failed label removal is silently ignored, which means `update_status` can add a new `status:*` label without removing the old one, exactly recreating the label sprawl described in PLAN.md.  
**Fix:** Distinguish 404 (label does not exist — safe to ignore) from other HTTP errors (5xx, rate limit — should propagate). Use the `gh api` exit code or stderr message to differentiate.

**3. `update_status` is a non-atomic read-modify-write**  
`src/backends/github.rs:47-64`: fetches labels, then loops over them calling `remove_label` per label, then adds the new label. Between `get_task` and the first `remove_label`, another concurrent call can race. This is inherent to the GitHub label API, but the `update_status` should at minimum document this, and consider using `gh api PATCH /repos/{repo}/issues/{number}` with the full desired label set in one call to reduce the window.

**4. Script-path discovery uses OsString lexicographic sort — wrong for semver, broken on Intel Macs**  
`src/engine/runner.rs:57-66`:
```rust
entries.filter_map(|e| e.ok()).max_by_key(|e| e.file_name())
```
`OsString::max` is lexicographic, so `0.53.9` > `0.53.10` (because `"9" > "1"`), returning the wrong version. Additionally, the path `/opt/homebrew/Cellar` only exists on Apple Silicon; Intel Macs use `/usr/local/Cellar`. The `HOMEBREW_CELLAR` or `HOMEBREW_PREFIX` environment variables (or `brew --cellar` output) should be used instead of a hardcoded path.  
**Fix:** Use `ORCH_SCRIPTS_DIR` as the primary path (already supported), and for the brew fallback call `brew --prefix orchestrator` via a subprocess or read `HOMEBREW_PREFIX`.

---

### Medium

**5. `gh api -f body=...` is not safe for multi-line or special-character content**  
`src/github/cli.rs:100-112` and `src/github/cli.rs:138-143`: The `-f field=value` syntax passes values through the shell argument vector, not a quoted JSON body. For issue bodies that contain newlines, backslashes, or embedded `=` signs, `gh api -f` may truncate or mangle the value. Use `--input -` with a JSON payload piped to stdin, or use `-F` (raw field) and let `gh` handle encoding.  
Example safer approach:
```rust
let payload = serde_json::json!({"title": title, "body": body, "labels": labels});
// pipe payload.to_string() to gh api --input -
```

**6. `ChannelRegistry` is instantiated but never populated**  
`src/engine/mod.rs:172`: `let _channels = ChannelRegistry::new();` creates the registry but nothing is ever registered into it. The underscore prefix suppresses the warning but hides that all channel infrastructure in this PR is entirely dead (Telegram, Discord, GitHub channel, tmux bridge). This is acceptable if the intent is structural scaffolding, but it should be clearly documented as Phase 3+ work, and the placeholder `bail!` implementations in `telegram.rs`/`discord.rs` mean accidentally calling any channel method panics the service.

**7. `drop(transport)` in `serve()` is a no-op**  
`src/engine/mod.rs` near the end of `serve()`: `drop(transport)` is explicit but does nothing beyond what Rust's drop-at-end-of-scope already does. Remove it.

**8. `jobs` table uses `type` as a column name (SQL reserved keyword)**  
`src/db.rs:114`: `type TEXT NOT NULL DEFAULT 'task'` — `type` is a reserved keyword in SQL. SQLite accepts it unquoted but it is fragile across query builders and migrations. Rename to `job_type`.

**9. Config `get()` re-reads disk on every call — no in-memory caching**  
`src/config/mod.rs:30-49`: The PLAN.md lists "In-memory config, hot-reload" as a Phase 1 deliverable, but the implementation reads the YAML file on every `get()` call. During a busy tick this adds unnecessary latency. At minimum, cache the parsed YAML in a `OnceLock<serde_yml::Value>` or a `RwLock<>` with a reload trigger.

---

### Minor

**10. `cargo clippy -- -D warnings -A dead_code -A unused`**  
`.github/workflows/release.yml:32`: Suppressing `-A dead_code -A unused` allows unimplemented placeholder code to accumulate undetected. During active development this is understandable, but track it as a tech debt item.

**11. Default new-repo version bumped from `v0.1.0` to `v1.0.0`**  
`.github/workflows/release.yml:163`: If the repo has no previous tags (fresh install), the first release will be tagged `v1.0.0`, but `Cargo.toml` says `1.0.0-alpha.1`. Consider aligning these or starting from `v0.1.0` as before and letting the version bump happen naturally.

**12. Cargo.lock is committed — fine for a binary crate, but `Cargo.lock` version = 4 requires Cargo ≥ 1.78**  
Ensure the CI pins a recent enough Rust stable (dtolnay/rust-toolchain@stable handles this, but worth noting for contributors running older toolchains locally).

**13. Formula `sha256` is all-zeros**  
`Formula/orch.rb:5`: `sha256 "000...000"` — this is clearly a placeholder that will be updated in CI at release time, but it means the formula cannot be tested locally until then. The existing `test` block should still pass since it comes from `head` installs, but document this limitation in `caveats` or a comment.

---

### Summary

The architecture is well-thought-out (incremental Rust migration, keeping bash for agent invocation, `ExternalBackend` trait for future trackers). The main blocker is items 1-4: the gitleaks exclusion without visible justification, the silent label-removal error masking `update_status` correctness, and the broken script-path lookup on Intel Macs. Items 5-9 should also be addressed before this ships. Items 10-13 are non-blocking.

---

gabrielkoerich:

The PR is a solid architectural foundation with a well-thought-out plan, but there are several issues that need to be addressed before merging.

## Critical Issues

### 1. Release binary only built for Linux — formula will be broken
`.github/workflows/release.yml:182-184` — `cargo build --release` runs on `ubuntu-latest`, producing a Linux x86_64 binary. The Homebrew formula (`Formula/orch.rb`) is for macOS users. Without cross-compiling for `aarch64-apple-darwin` and `x86_64-apple-darwin`, the release artifact won't be usable on any Mac. The release job needs `cargo build --target aarch64-apple-darwin` and `x86_64-apple-darwin` steps, or a matrix build, plus `lipo` to produce a universal binary (as described in PLAN.md's Brew Upgrade Path section but not implemented in the workflow).

### 2. `remove_label` silently swallows errors — can corrupt task state
`src/github/cli.rs:131-134` — The function logs a warning but always returns `Ok(())` even on non-404 errors. In `src/backends/github.rs:52-61`, `update_status` calls `remove_label` for every existing `status:*` label before adding the new one. If any removal silently fails, the task ends up with multiple `status:*` labels, which will cause `list_by_status` to return tasks in unexpected states (e.g., a task appearing in both `in_progress` and `done`). The fix should only swallow 404 errors (label already removed) and propagate all other errors:
```rust
pub async fn remove_label(...) -> anyhow::Result<()> {
    let endpoint = ...; 
    match self.api(&[&endpoint, "-X", "DELETE"]).await {
        Ok(_) => Ok(()),
        Err(e) if e.to_string().contains("404") => Ok(()), // already removed
        Err(e) => Err(e),
    }
}
```

### 3. Hardcoded Homebrew path fails on Intel Macs
`src/engine/runner.rs:68` — The scripts directory discovery hardcodes `/opt/homebrew/Cellar` (Apple Silicon Homebrew prefix). Intel Macs use `/usr/local/Cellar`. The fallback to `orch_home.join("scripts")` may save it in practice, but the primary path is wrong for ~30% of Mac users. Use `brew --prefix` or the `HOMEBREW_PREFIX` env var, or rely solely on `ORCH_SCRIPTS_DIR` env var and the `PATH`-based fallback.

### 4. Formula SHA256 is all zeros
`Formula/orch.rb:5` — `sha256 "000...000"` is a placeholder. This formula cannot be installed. The release workflow needs to compute the real SHA256 after building the tarball and substitute it. If this is intentional as a placeholder for the tap-update step, ensure the `sed` substitution in the release workflow (`.github/workflows/release.yml:269-272`) actually hits this new formula file — it does reference `Formula/orch.rb` now, which is correct.

## Correctness Issues

### 5. `update_status` is non-atomic
`src/backends/github.rs:51-65` — The pattern get-current-labels → remove-old → add-new involves multiple API calls with no locking. Two concurrent engine ticks could both read the same labels, both remove the old status, and both try to add different new statuses, resulting in an incorrect final state. Consider using the GitHub API's direct label endpoint with a replace-all approach, or accept the race as a known limitation with a comment.

### 6. Clippy allows `dead_code` and `unused`
`.github/workflows/release.yml:30` — `cargo clippy -- -D warnings -A dead_code -A unused` masks legitimate warnings from the many stub/placeholder implementations (discord, telegram, stream_output TODOs, sync_tick TODOs). These should emit warnings to track technical debt. Either implement them minimally (returning structured `Err`) or keep the lints enabled and suppress them per-item with `#[allow(dead_code)]` on specific stub structs so the suppression is visible and intentional.

### 7. Intel-incompatible path also affects `project_dir`
`src/engine/runner.rs:107-121` — `project_dir()` uses `home.join("Projects")` as a fallback, but this is a personal convention not a system standard. The final fallback to `current_dir()` is reasonable but may resolve to the Homebrew Cellar directory when launched as a service, not the user's project. Consider requiring `PROJECT_DIR` to be set in config rather than guessing.

## Minor Issues

### 8. Version bump to `v1.0.0` for first-release projects is aggressive
`.github/workflows/release.yml:166` — Previously `v0.1.0`, now `v1.0.0` for repos with no prior tags. This contradicts the `1.0.0-alpha.1` version in `Cargo.toml`. Should start at `v0.1.0` or `v1.0.0-alpha.1` consistently.

### 9. Config `get()` reads from disk on every call
`src/config/mod.rs:25-40` — One of the stated performance goals is eliminating repeated `yq` subprocesses for config reads. The current implementation still reads and parses the YAML file on every call to `get()`. For a service that ticks every 10s, this should at minimum cache the parsed config in a `once_cell::sync::OnceCell` or equivalent, with hot-reload support on file change.

### 10. `db.default_path()` uses `.orchestrator` but PLAN.md targets `.orch`
`src/db.rs:14` — The path `~/.orchestrator/orchestrator.db` is the v0 convention. PLAN.md's Name Change section says the v1 target is `~/.orch/`. This inconsistency should be decided before merging to avoid a migration headache later.

### 11. Diff truncated — `src/main.rs` CLI dispatch not reviewable
The diff is truncated at 5000 lines. `src/main.rs` only shows module declarations (`mod backends; mod channels; ... mod security;`). The CLI command dispatch (clap subcommands for `serve`, `config get`, `cron check`, etc.) is not visible. This is a significant gap in reviewability. Similarly `src/parser.rs` and `src/security.rs` are not included in the diff.

---

gabrielkoerich:

This is a substantial, well-planned PR introducing the Rust core. The architecture, trait design, and test coverage are solid. Several issues need to be addressed before merging:

**Blocking Issues:**

1. **`.gitleaks.toml:12` — `security.rs` excluded from secret scanning without justification.** The diff adds `src/security.rs` to the gitleaks ignore paths, but that file is not visible in the diff (truncated at 5000 lines). Excluding a file named `security.rs` from secret scanning requires explicit justification. If it contains test patterns or known-fake credentials, that needs a comment in `.gitleaks.toml`. This should be explained or the exclusion removed.

2. **`src/engine/jobs.rs:63–87` — Only `status:done` clears `active_task_id`; all other statuses silently skip re-scheduling.** If a task ends in `status:needs_review`, `status:blocked`, or `status:in_review`, the `active_task_id` is never cleared, and the job will never fire again until the user manually edits `jobs.yml`. Fix: treat any terminal or non-in-progress status as clearable. At minimum, add `status:needs_review` and `status:failed` to the match arm that clears `active_task_id`.

3. **`src/engine/runner.rs:112–134` — `project_dir()` heuristic is wrong for most users.** The method checks `~/Projects/{repo_name}`, which is a personal convention unlikely to match other installs. This would cause `run_task.sh` to be invoked with the wrong `PROJECT_DIR` silently (falls back to cwd). This should either be read from config (`config::get("project_dir")`) or documented prominently as requiring explicit configuration.

4. **`Formula/orch.rb:48` — `--version` flag may not be implemented.** The test does `shell_output("#{bin}/orch-core --version 2>&1", 0)`, but `src/main.rs` is not visible in the truncated diff. If clap's `version` attribute is not set in `Cargo.toml` or the binary subcommand, this test will fail. Verify `clap` is configured with `.version()` or the `version` feature.

**Non-Blocking But Should Fix:**

5. **`.github/workflows/release.yml:24` — `cargo clippy` suppresses `dead_code` and `unused` warnings for the entire codebase.** `cargo clippy -- -D warnings -A dead_code -A unused` is too broad. Placeholder channel implementations (Telegram, Discord) have obviously dead structs, but `TmuxChannel`, `ChannelRegistry`, and `Transport` fields could also be silently unused. Prefer `#[allow(dead_code)]` on specific placeholder items rather than a global flag.

6. **`src/backends/github.rs:46–57` — `update_status` has a TOCTOU race.** The method calls `get_task` to fetch current labels, then removes status labels one-by-one, then adds the new one — three separate API operations. Concurrent task runners could interleave these. Consider using `PATCH /repos/{repo}/issues/{number}` with the full label set, or at minimum note this in a comment since the `gh` CLI path makes true atomicity difficult here.

7. **`src/channels/transport.rs:83–98 and 101–113` — `bind()` and `unbind()` take two separate async write locks non-atomically.** If a task is cancelled between the first and second lock acquisition in `bind()`, `bindings` and `thread_to_task` can get out of sync. Consider combining them into a single `Mutex<(HashMap, HashMap)>` or using a single lock covering both maps.

8. **`src/config/mod.rs:23–38` — `get()` re-reads and re-parses YAML from disk on every call.** The engine tick calls config reads frequently (routing, job dispatch). For Phase 1 this is tolerable, but the function should at minimum be documented as non-cached. The PLAN.md describes in-memory config with hot-reload — even a simple `OnceLock` or `RwLock<Value>` would be worthwhile here.

9. **`src/db.rs:96–120` — No schema versioning.** Using `PRAGMA user_version` would make future migrations safe. Right now, `IF NOT EXISTS` means schema changes on upgrade are silently skipped for existing installations. Add at minimum a `PRAGMA user_version` check.

10. **`release.yml:180–183` — Release CI builds on `ubuntu-latest` only.** The binary produced is Linux/x86_64. The Homebrew formula correctly builds from source (`depends_on "rust" => :build`), so this is fine for now. However, the PLAN.md explicitly describes prebuilt macOS binaries (aarch64 + x86_64) as part of the release flow. The current release job should either be updated to match that plan or the formula's `install` method and PLAN.md should be aligned. At minimum, note this gap.

**Note:** The diff is truncated at 5000 lines, so `src/main.rs`, `src/parser.rs`, `src/sidecar.rs`, `src/tmux.rs`, and `src/security.rs` are not visible. The review of the `--version` flag (item 4) and the `security.rs` exclusion (item 1) depend on those files.

---

PR #311, opened 2026-02-23
fix: multi-project polling and per-project jobs

gabrielkoerich:
## Summary
- **Multi-project poll**: `serve.sh` now iterates `projects.yml` for `poll.sh` on every tick (not just cleanup/review/mentions), so one service manages multiple repos
- **Per-project jobs**: `jobs.yml` lives at `PROJECT_DIR/.orchestrator/jobs.yml` (falls back to `ORCH_HOME/jobs.yml` for backward compat)
- **Project config loading**: `poll.sh` and `jobs_tick.sh` call `load_project_config` so `_GH_REPO` and other settings resolve for the correct project
- **Project discovery**: `db_task_projects()` reads from `projects.yml` registry in addition to sidecar data

## Test plan
- [x] All 272 bats tests pass
- [ ] Restart service, verify both orchestrator and orch repos are polled
- [ ] Create a task on orch repo, verify orchestrator picks it up
- [ ] Create a project-local jobs.yml, verify it fires on schedule

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,6 @@ name: CI
 
 on:
   push:
-  pull_request:
   workflow_dispatch: {}
 
 permissions:


gabrielkoerich:

The PR's multi-project polling approach is sound and the per-project `jobs.yml` lookup logic in `_jobs_file()` is correct. However there is one concrete bug and a few issues worth addressing.

---

## Bug: global jobs run in the wrong project context when a project has an `orchestrator.yml` but no local `jobs.yml`

**Affected files:** `scripts/serve.sh` (new per-project loop), `scripts/jobs_tick.sh` (new `load_project_config` call at line 7), `scripts/backend.sh` `_jobs_file()` fallback.

**Scenario:** Project A is in `projects.yml`, has an `orchestrator.yml` (so `_GH_REPO`, etc. are set to A's repo), but has no `.orchestrator/jobs.yml`.

1. `serve.sh` calls `PROJECT_DIR=/path/to/A jobs_tick.sh`.
2. `load_project_config` finds A's `orchestrator.yml`, sets `STATE_DIR` → `A/.orchestrator`, `CONFIG_PATH` → A's merged config, `_GH_REPO` → A's GitHub repo.
3. `_jobs_file()` returns `ORCH_HOME/jobs.yml` (correct fallback) — so the right jobs file is read.
4. A task-type job fires. It calls `add_task.sh` with `PROJECT_DIR="$JOB_DIR"` (line 133 of `jobs_tick.sh`), which is correct for task creation.
5. **Bug:** immediately after, `db_task_set "$NEW_TASK_ID" "agent" "$JOB_AGENT"` (line 138) executes in the outer shell where `_GH_REPO` is now set to A's repo. If A's repo differs from `JOB_DIR`'s repo, this GitHub API call targets the wrong repository and will either fail silently or update the wrong task.
6. `JOBS_LOG` (line 10 of `jobs_tick.sh`) is set to `A/.orchestrator/jobs.log` instead of the global log, so global job output is scattered across project state dirs.

The minute-level dedup at `jobs_tick.sh:62-68` prevents double-firing, but only the first project alphabetically "owns" the execution context — which project that is is non-deterministic (depends on `sort -u` output).

**Suggested fix:** In `serve.sh`, run the global `jobs_tick.sh` (without `PROJECT_DIR`) once per tick regardless, and only run a per-project invocation when that project actually has a local `jobs.yml`:

```sh
# Global tick always runs once
"$SCRIPT_DIR/jobs_tick.sh" >> "$LOG_FILE" 2>&1 || true
# Per-project tick only when a local jobs file exists
if [ -n "$POLL_DIRS" ]; then
  while IFS= read -r pdir; do
    [ -f "$pdir/.orchestrator/jobs.yml" ] || continue
    PROJECT_DIR="$pdir" "$SCRIPT_DIR/jobs_tick.sh" >> "$LOG_FILE" 2>&1 || true
  done <<< "$POLL_DIRS"
fi
```

---

## Minor: `JOBS_FILE` global var is now dead weight

`scripts/backend.sh` line after `_jobs_file()` still sets:
```sh
JOBS_FILE="${JOBS_FILE:-${ORCH_HOME}/jobs.yml}"
```
`_jobs_file()` inlines the same fallback expression. If no other code references `$JOBS_FILE` directly, this variable assignment can be removed to avoid confusion. If it IS referenced elsewhere, `_jobs_file()` should not re-implement the fallback — it should just call `echo "$JOBS_FILE"`.

---

## Minor: `yq -r` in `db_task_projects` is inconsistent with the rest of the file

`scripts/backend_github.sh` (new code inside `db_task_projects`):
```sh
yq -r '.projects[]?.path // empty' "$registry" 2>/dev/null || true
```
Every other `yq` call in `backend_github.sh` uses `-o=json` (e.g. `yq -o=json '.jobs // []'`). The `-r` flag works with go-yq v4 for raw string output, but it is worth confirming the installed `yq` version handles `// empty` the same way jq does. The `2>/dev/null || true` silences any failure, so a version incompatibility would silently return no projects, causing all polling to be skipped until the fallback kicks in.

---

## Note: no automated test coverage for `projects.yml` registry path

The PR description lists the registry-based discovery as a new code path in `db_task_projects`, but the test-plan checkboxes for multi-project integration are all manual. It would be worth adding a bats test that creates a `projects.yml`, calls `db_task_projects`, and asserts the registered path is returned.

---

PR #317, opened 2026-02-23
fix: include project name in tmux session names to avoid collisions

### LINKED ISSUE #312: tmux session names should include project name to avoid collisions
## Problem

tmux session names use `orch-{issue_number}` format (`run_task.sh` line 528):

```bash
TMUX_SESSION="orch-${TASK_ID}"
```

With multi-project polling, two different projects can have GitHub issues with the same number (e.g. both have issue #5). This causes tmux session name collisions — the second task would kill the first task's tmux session.

## Fix

Include `PROJECT_NAME` (already available on line 227) in the session name:

```bash
TMUX_SESSION="orch-${PROJECT_NAME}-${TASK_ID}"
```

`PROJECT_NAME` is derived from `basename "$PROJECT_DIR" .git`, so for `~/Projects/my-app` it becomes `my-app`, giving session names like `orch-my-app-5`.

## Files

- `scripts/run_task.sh` line 528 — session name
- `scripts/run_task.sh` — any `tmux kill-session` or attach references
- Tests referencing `orch-` session pattern

gabrielkoerich:
## Summary

- Include `PROJECT_NAME` in tmux session names: `orch-{project}-{task_id}` instead of `orch-{task_id}`
- Update `task_attach.sh` and `task_kill.sh` to find sessions by suffix pattern matching
- Update `task_live.sh` to correctly extract task ID from the new format

## Problem

With multi-project polling, two projects can have GitHub issues with the same number (e.g., both have issue #5). The old `orch-{task_id}` format caused session name collisions — the second task's `tmux new-session` would kill the first task's session.

## Fix

`PROJECT_NAME` (derived from `basename "$PROJECT_DIR" .git`) is already set when `TMUX_SESSION` is assigned. Changed to `orch-${PROJECT_NAME}-${TASK_ID}` for unique cross-project session names (e.g. `orch-my-app-5` and `orch-other-app-5`).

The helper scripts use `tmux list-sessions | grep -E "^orch-.*-${TASK_ID}$"` to find sessions by task ID suffix — no project context required when calling `orch task attach <id>` or `orch task kill <id>`. `task_live.sh` uses `${name##*-}` to strip the variable-length project prefix.

## Test plan

- [x] Bash syntax check passes for all 4 modified files (`bash -n`)
- [x] No existing tests reference tmux session naming
- Manual: run two tasks from different projects with same issue number and verify no collision

Closes #312

### DIFF
--- a/scripts/run_task.sh
+++ b/scripts/run_task.sh
@@ -512,7 +512,7 @@ ASYNC_TMUX=${ASYNC_TMUX:-$(config_get '.workflow.async_tmux // "true"')}
 _USED_TMUX=false
 TMUX_RESPONSE_FILE="${STATE_DIR}/${FILE_PREFIX}-tmux-response-${ATTEMPTS}.txt"
 TMUX_STATUS_FILE="${STATE_DIR}/${FILE_PREFIX}-tmux-status-${ATTEMPTS}.txt"
-TMUX_SESSION="orch-${TASK_ID}"
+TMUX_SESSION="orch-${PROJECT_NAME}-${TASK_ID}"
 USE_TMUX=${USE_TMUX:-$(config_get '.workflow.use_tmux // "true"')}
 
 # Collect agent output and update task. Called after agent finishes (sync or async).
--- a/scripts/task_attach.sh
+++ b/scripts/task_attach.sh
@@ -10,17 +10,25 @@ if [ -z "$TASK_ID" ]; then
   exit 1
 fi
 
-SESSION="orch-${TASK_ID}"
-
 if ! command -v tmux >/dev/null 2>&1; then
   log_err "tmux is not installed"
   exit 1
 fi
 
-if ! tmux has-session -t "$SESSION" 2>/dev/null; then
-  log_err "No active tmux session for task $TASK_ID (session: $SESSION)"
+# Session names include project: orch-{project}-{task_id}. Search by task ID suffix.
+_SESSIONS=$(tmux list-sessions -F '#{session_name}' 2>/dev/null \
+  | grep -E "^orch-.*-${TASK_ID}$" || true)
+SESSION=$(printf '%s' "$_SESSIONS" | head -1)
+
+if [ -z "$SESSION" ]; then
+  log_err "No active tmux session for task $TASK_ID"
   log_err "The agent may have already finished. Check: orch task live"
   exit 1
+fi
+
+_SESSION_COUNT=$(printf '%s\n' "$_SESSIONS" | grep -c . || true)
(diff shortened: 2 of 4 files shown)


gabrielkoerich:

The PR correctly solves the stated collision problem. The session naming change is sound, pattern matching logic works, and `task_live.sh`'s `${name##*-}` correctly extracts the task ID regardless of how many hyphens the project name contains. A few concerns worth noting:

**1. PROJECT_NAME may contain tmux-invalid characters (scripts/run_task.sh:515)**
Tmux session names cannot contain spaces, colons, or some special characters. `PROJECT_NAME` is derived from `basename "$PROJECT_DIR"`, so a directory named `my app` or `my:project` would cause `tmux new-session` to fail silently or error. Consider sanitizing: `PROJECT_NAME_SAFE=$(printf '%s' "$PROJECT_NAME" | tr -cs 'a-zA-Z0-9_-' '-')` and using that in the session name.

**2. TASK_ID used unescaped in ERE pattern (scripts/task_attach.sh:18, scripts/task_kill.sh:18)**
`grep -E "^orch-.*-${TASK_ID}$"` is fine for GitHub issue numbers (always numeric), but TASK_ID is sourced from user input. If it ever contained ERE metacharacters (`.`, `*`, `?`, `[`), the pattern could match unintended sessions. Low practical risk given numeric IDs, but worth noting.

**3. task_kill.sh only kills first of multiple matching sessions (scripts/task_kill.sh:31-36)**
When `_SESSION_COUNT -gt 1`, a warning is emitted and only the first session is killed. The remaining sessions are left running. This could be intentional, but it seems like a footgun — the user calls `orch task kill <id>` and thinks the task is dead while a second session continues. Consider either killing all matched sessions or making the multi-session case a hard error requiring explicit project-qualified kill.

**4. Backward compatibility with in-flight sessions**
Existing sessions in the old `orch-{TASK_ID}` format won't match the new `^orch-.*-{TASK_ID}$` pattern in `task_attach.sh` and `task_kill.sh`, because the pattern requires at least one hyphen-separated component between `orch-` and the task ID. Sessions created before this deploy will be invisible to the helper scripts until they expire. This is expected for a naming-scheme change, but should be called out in upgrade/deployment notes.

**5. `task_live.sh` - no functional issue, but filter scope**
`${name##*-}` on the old format `orch-5` still yields `5`, so `task_live.sh` remains backward compatible with pre-existing sessions. No issue here.

Overall the implementation is correct for the described scenario. Items 1 and 3 are the most actionable: sanitizing `PROJECT_NAME` prevents a hard-to-debug failure mode, and silently leaving orphaned sessions on a multi-kill is surprising behavior.

---

PR #374, opened 2026-02-25
fix(lib): correct pick_fallback_agent start index logic

### LINKED ISSUE #364: Bug: pick_fallback_agent has incorrect start index when current agent not found
## Summary
The `pick_fallback_agent` function in `scripts/lib.sh` (lines 416-451) has a bug in the start index calculation when the current agent is not found in the available agents list.

## Root Cause
```bash
# Line 426-434
local n="${#list[@]}"
# Default start so that when current is not found we begin from list[0].
local start=$(( n - 1 ))   # Line 428: Bug!
for i in "${!list[@]}"; do
  if [ "${list[$i]}" = "$current" ]; then
    start="$i"
    break
  fi
done
```

The comment says "when current is not found we begin from list[0]", but the code sets `start=$(( n - 1 ))` which is the **last** element. Combined with the loop below:

```bash
# Line 437-439
for offset in $(seq 1 "$n"); do
  idx=$(((start + offset) % n))
```

When `start = n-1` and `offset = 1`, the first candidate checked is `list[0]` (which is correct by coincidence). However, the logic is confusing and fragile.

More critically: if `current` IS found at index 0, then `start=0`, and the first offset iteration gives `idx=1`, which **skips** the first agent entirely when current is at position 0!

## Symptom
When trying to pick a fallback agent, the function may skip the first available agent or have incorrect behavior depending on where the current agent is in the list.

## Fix Required
Simplify the logic - just iterate from the next agent after current:
```bash
for i in "${!list[@]}"; do
  if [ "${list[$i]}" = "$current" ]; then
    # Start from next agent
    for offset in $(seq 1 "$n"); do
      idx=$(((i + offset) % n))
      # check candidate...
    done
    return
  fi
done
# current not found - start from first agent
for agent in "${list[@]}"; do
  # check agent...
done
```

## Files Affected
- `scripts/lib.sh:416-451` - `pick_fallback_agent` function

gabrielkoerich:
## Summary
- Fixed `pick_fallback_agent` in `scripts/lib.sh` which had an incorrect start index when the current agent was not found in the available agents list
- The old code set `start=$(n-1)` which worked by coincidence but was fragile and misleading
- Refactored into two clear branches: iterate from next agent when current is found, iterate from first when not
- Added 8 comprehensive tests covering all edge cases

## Test plan
- [x] All 8 new `pick_fallback_agent` tests pass
- [x] Full test suite (294 tests) passes with no regressions

Closes #364

### DIFF
--- a/docs/content/posts/evening-retrospective-2026-02-25.md
+++ b/docs/content/posts/evening-retrospective-2026-02-25.md
@@ -1,0 +1,127 @@
++++
+title = "Evening Retrospective — 2026-02-25"
+date = "2026-02-25"
++++
+
+## Evening Retrospective — 2026-02-25
+
+### Today's Fixes (from git log)
+
+| Commit | Description |
+|--------|-------------|
+| `8afe609` | fix(review): match run_task.sh opencode execution exactly |
+| `4fd123d` | fix(review): write prompt to temp file for opencode |
+| `b6e81e6` | fix(review): use correct opencode CLI syntax |
+| `0296cfe` | fix(review): add more logging for agent failures |
+| `d16f261` | fix: round-robin review agent with task exclusion + skills.yml lookup |
+| `9e81fd2` | fix(run_task.sh): fix retry loop detection subshell variable scoping bug (#357) |
+| `012e31a` | fix(security): validate job commands before execution (#350) |
+| `c215035` | fix(reliability): preserve malformed agent responses for debugging (#349) |
+
+**Major improvements today:**
+- **Review agent opencode fixes** — Four commits addressing opencode execution in review mode (temp files, CLI syntax, execution matching)
+- **Retry loop detection fixed** — PR #357 merged, fixing the subshell variable scoping bug
+- **Security hardening** — Job command validation before execution
+- **Debuggability** — Malformed agent responses now preserved for analysis
+
+---
+
+### Current Task Status
+
+| Status | Count |
+|--------|-------|
+| new | 1 |
+| routed | 0 |
+| in_progress | 2 |
+| needs_review | 4 |
+| done | 55 |
+
+**In Progress:**
+- **#370** (kimi) — This evening retrospective task
+- **#368** (minimax) — Code quality: Inconsistent return code patterns — **stuck for 1904s earlier, recovered by poll**
+
+**Needs Review (bug reports ready for fix):**
+- **#364** (kimi) — `pick_fallback_agent` has incorrect start index when current agent not found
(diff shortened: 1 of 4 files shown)


gabrielkoerich:
## 🔍 Automated Review — Approve

## Overall Assessment

The refactoring is correct and the tests are a valuable addition. A few observations:

### Correctness: Looks good

Both branches of the new code are logically correct:

- **Found branch** (`scripts/lib.sh`): Uses the same modular arithmetic as before (`(current_idx + offset) % n`) starting from offset=1, correctly skipping the current agent and any excluded agents.
- **Not-found branch** (`scripts/lib.sh`): Straightforward linear iteration from the beginning of the list, correctly filtering out excluded agents.

The removal of the `[ -n "$current" ]` guard in the found branch (line ~448 in the new code) is safe because `current_idx` being non-empty guarantees `current` is non-empty.

### Accuracy of PR description

One minor note: the PR describes this as fixing an "incorrect start index" and the old code as working "by coincidence." In reality, `start=$(( n - 1 ))` with `offset` starting at 1 is a standard modular arithmetic pattern — `(n-1+1) % n = 0` — which correctly starts iteration from index 0. The original code was **functionally correct** in all cases, not broken. This is really a **readability refactor**, not a bug fix. The new two-branch structure is genuinely clearer about intent, which is a worthwhile improvement, but calling it a fix is slightly misleading.

### Tests

The 8 new tests (`tests/orchestrator.bats`, lines after 5771) provide good coverage:
- Basic next-agent selection
- Wrap-around behavior
- Exclusion filtering
- Current-not-found fallback
- Exclusion when current not found
- All-excluded failure case
- Empty current string

**Minor nit:** Test 7 ("with current at index 0 does not skip first agent") is functionally identical to test 1 — same inputs (`current=claude`, agents=`claude,codex,opencode`), same expected output (`codex`). The comment references an "original bug" but since the original code handled index 0 correctly, this test is redundant. Not a blocker, just noting it.

### Code duplication

The candidate filtering logic (empty check, current check, exclude check) is now duplicated across both branches. This is a reasonable trade-off for readability but worth noting as a future maintenance consideration if the filtering logic ever needs to change.

### Trailing whitespace

The diff shows a trailing space on the blank line between the last new test and the existing `run_with_timeout` test (line ~5860 in the new test file). Very minor, some CI linters may flag it.

### Verdict

The code is correct, the tests are comprehensive, and the refactoring genuinely improves readability. Approving.

---
*By kimi[bot] using model `opus` via [Orchestrator](https://github.com/gabrielkoerich/orchestrator)*

PR #403, opened 2026-02-27
fix: cleanup_worktrees silently skips when gh unavailable

### LINKED ISSUE #401: fix: cleanup_worktrees silently skips when gh unavailable
## Summary

In scripts/cleanup_worktrees.sh line 88, the `issue_has_merged_pr()` function silently continues if gh is unavailable or the command fails:

```bash
if ! issue_has_merged_pr "$repo" "$gh_issue"; then
  continue
fi
```

When gh is not available or the API call fails, the function returns 1 (failure), causing the script to `continue` and skip worktree cleanup entirely.

## Root Cause

The function checks for merged PRs but doesn't distinguish between:
1. gh not installed
2. API call failed (rate limit, network error)
3. No merged PR found (correctly skip cleanup)

This can leave orphaned worktrees behind when gh is temporarily unavailable.

## Fix

Add explicit error handling:
```bash
if ! command -v gh >/dev/null 2>&1; then
  log "[cleanup_worktrees] gh not available, skipping"
  continue
fi

if ! issue_has_merged_pr "$repo" "$gh_issue"; then
  if [ $? -eq 2 ]; then  # gh error, not "no PR found"
    log_err "[cleanup_worktrees] failed to check PR status"
    continue
  fi
  # No merged PR found - continue (correct behavior)
fi
```

Or add a separate return code for "gh unavailable" vs "no merged PR".

## Files Affected

- scripts/cleanup_worktrees.sh:88

gabrielkoerich:
Closes #401

### DIFF
--- a/scripts/cleanup_worktrees.sh
+++ b/scripts/cleanup_worktrees.sh
@@ -45,8 +45,12 @@ issue_has_merged_pr() {
   local issue_number="$2"
   local pr_number=""
 
+  # Return codes:
+  # 0 = merged PR found
+  # 1 = no merged PR found
+  # 2 = gh unavailable or API error
   if ! command -v gh >/dev/null 2>&1; then
-    return 1
+    return 2
   fi
 
   pr_number=$(gh pr list \
@@ -54,8 +58,16 @@ issue_has_merged_pr() {
     --state merged \
     --search "closes #$issue_number" \
     --json number \
-    --jq '.[0].number // ""' 2>/dev/null || true)
-  [ -n "$pr_number" ]
+    --jq '.[0].number // ""' 2>/dev/null) || {
+    # gh command failed (network, auth, rate limit, etc.)
+    return 2
+  }
+
+  if [ -n "$pr_number" ]; then
+    return 0
+  fi
+
+  return 1
 }
 
 log "[cleanup_worktrees] [$PROJECT_NAME] scan start"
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
## 🔍 Automated Review — Changes Requested

## Bug: `$?` is always 0 inside the `if ! ...; then` block

The core fix in this PR has a critical logic bug at **`scripts/cleanup_worktrees.sh` lines 100-112**.

When the shell evaluates:
```bash
if ! issue_has_merged_pr "$repo" "$gh_issue"; then
    case "$?" in
```

The `!` negation means that if `issue_has_merged_pr` returns non-zero (1 or 2), `! issue_has_merged_pr` evaluates to **0** (true), and we enter the `then` block. At that point, `$?` is **always 0** — it reflects the exit status of the entire `if` condition, not the original return code from the function.

This means **none of the `case` branches (1, 2, *) will ever match**, making the new logging dead code. The `gh unavailable` error will never be surfaced, which defeats the stated purpose of this PR.

### Suggested fix

Capture the return code before the conditional test:

```bash
issue_has_merged_pr "$repo" "$gh_issue"
rc=$?
if [ "$rc" -ne 0 ]; then
    case "$rc" in
      1)
        log "[cleanup_worktrees] [$PROJECT_NAME] task=$id no merged PR; skipping"
        ;;
      2)
        log_err "[cleanup_worktrees] [$PROJECT_NAME] task=$id failed to check PR status for repo=$repo issue=$gh_issue; gh unavailable or API error; skipping cleanup for now"
        ;;
      *)
        log_err "[cleanup_worktrees] [$PROJECT_NAME] task=$id issue_has_merged_pr returned unexpected code=$rc; skipping"
        ;;
    esac
    continue
fi
```

---

The changes to `issue_has_merged_pr` itself (lines 48-73) look correct — the three-way return code (0/1/2) and the explicit error handling for `gh` failures are a good improvement over the previous `|| true` swallowing. Just the caller side needs fixing.

---
*By minimax[bot] using model `opus` via [Orchestrator](https://github.com/gabrielkoerich/orchestrator)*

